PR #520, opened 2026-03-30
feat: Add formatting functions to SurveyJS

### LINKED ISSUE #492: Add custom expression functions for formatting (Currency, Percentages, Dates)
**User Story:**
As a survey creator, I want to be able to format variables inside my survey text (e.g., displaying a calculated total as "$1,200.00" instead of "1200") so that the final output looks professional, localized, and easy to read.

**Current Behavior:**
When using variables in text or expressions (e.g., `{total_score}` or `{price}`), SurveyJS outputs the raw numeric or string value. There is no built-in way to easily format these values as currencies, percentages, or localized dates directly within the expression engine.

**Expected Behavior:**
We should register a set of custom functions to the SurveyJS `FunctionFactory` that authors can use inside expressions to format data. 

**Proposed Implementation:**
Leverage the native JS `Intl` API (`Intl.NumberFormat` and `Intl.DateTimeFormat`) to keep the implementation lightweight and avoid importing heavy formatting libraries.

Register the following custom functions via `Survey.FunctionFactory.Instance.register()`:

1. **`formatCurrency(value, currencyCode, locale)`**
   * *Usage:* `formatCurrency({total_cost}, 'USD', 'en-US')` 
   * *Output:* `$1,234.56`
   * *Usage 2:* `formatCurrency({total_cost}, 'EUR', 'de-DE')`
   * *Output:* `1.234,56 €`
   
2. **`formatNumber(value, decimalPlaces, locale)`**
   * *Usage:* `formatNumber({calc_result}, 2, 'en-US')`
   * *Output:* `1,234.56` (handles thousand separators and fixed decimals)

3. **`formatDate(value, formatStyle, locale)`**
   * *Usage:* `formatDate({birthdate}, 'short', 'en-US')` -> `12/31/2023`
   * *Usage:* `formatDate({birthdate}, 'long', 'en-US')` -> `December 31, 2023`

**Acceptance Criteria:**
- [ ] Functions are registered and available in the SurveyJS expression runner.
- [ ] `formatCurrency` correctly applies the designated currency symbol and localized formatting.
- [ ] `formatNumber` correctly handles thousand separators and fixed decimal places.
- [ ] The functions fail gracefully (e.g., return the original raw value or an empty string) if the input value is null, undefined, or an invalid type.
- [ ] Add documentation/tooltips so users know what parameters to pass.

PeterMarinov:
# feat: Add formatting functions to SurveyJS

## Description
- Includes the following formatting functions - formatCurrency, formatNumber, formatDate, smartFormat
- Updates extensions' core registry to allow static imports of extensions so we have one place to add core and user extensions
- Adds utils for expressions to parse params
- Docs will be added in seperate Pr

Please include a summary of the changes and the related issue. Also, explain the context and the reason for the proposed changes.

To validate use this survey
```json
{
  "pages": [
    {
      "name": "orderPage",
      "elements": [
        {
          "type": "text",
          "name": "basePrice",
          "title": "Base Price",
          "defaultValue": 99.99
        },
        {
          "type": "radiogroup",
          "name": "country",
          "title": "Select Country",
          "choices": [
            {
              "value": "US",
              "text": "United States"
            },
            {
              "value": "UK",
              "text": "United Kingdom"
            },
            {
              "value": "DE",
              "text": "Germany"
            }
          ]
        },
        {
          "type": "expression",
          "name": "taxAmount",
          "title": "Tax {formattedTaxRate}",
          "description": "{readonlyTaxRate}",
          "expression": "{basePrice} * {taxRate}"
        },
        {
          "type": "expression",
          "name": "readonlyTaxRate",
          "visible": false,
          "expression": "{taxRate}",
          "format": "Tax rate from expression element is {0}",
          "displayStyle": "percent"
        },
        {
          "type": "expression",
          "name": "total",
          "title": "Total",
          "expression": "{basePrice} + {taxAmount}"
        },
        {
          "type": "expression",
          "name": "formattedTotal",
          "title": "Formatted Total",
          "expression": "formatCurrency({total})"
        },
        {
          "type": "expression",
          "name": "formattedDate",
          "title": "Order Date",
          "expression": "formatDate(today(), 'full', 'BG')"
        },
        {
          "type": "expression",
          "name": "question1",
          "title": "{formattedPrice}",
          "expression": "{formattedPrice}"
        }
      ]
    }
  ],
  "calculatedValues": [
    {
      "name": "taxRate",
      "expression": "iif({country} = 'US', 0.08, iif({country} = 'UK', 0.20, 0.19))"
    },
    {
      "name": "formattedTaxRate",
      "expression": "format({taxRate}, 'percent')"
    },
    {
      "name": "formattedPrice",
      "expression": "formatCurrency({basePrice}, \"GBP\", \"BG\")"
    }
  ]
}
```

## Related Issues
- closes https://github.com/endatix/endatix-hub/issues/492

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [x] Refactoring (non-breaking change which improves code quality)
- [ ] Other (please describe)

## Checklist

> [!NOTE]

## Screenshots
If applicable, add screenshots to help explain your changes.


lib/survey-features/expression-formatting/formatters.ts
```diff
@@ -0,0 +1,161 @@
...
+    }
+    case "percent": {
+      const num = parseNumberValue(valueParam);
+      if (num === null) {
+        return String(valueParam);
+      }
+      return new Intl.NumberFormat(undefined, { style: "percent" }).format(num);
+    }
+    case "date": {
+      const dataStyle = getStringParam(params, 3) ?? DEFAULT_DATE_STYLE;
```
orilux: nit: it should be `dateStyle`

Isn't param 3 is supposed to be locale? This doesn't align with the comment above the function's declaration.

lib/utils/formatters.ts
```diff
@@ -1,3 +1,19 @@
+type DateStyle = "full" | "long" | "medium" | "short";
+
+function isValidDate(value: unknown): value is Date {
+  return value instanceof Date && !Number.isNaN(value.getTime());
+}
+
+function parseNumberValue(value: unknown): number | null {
+  if (typeof value === "number") return value;
+  if (typeof value === "string") {
+    const parsed = Number(value);
```
orilux, line 27: If `value` is an empty string `Number(value)` will incorrectly return `0`. I recommend an explicit check similar to:

```
if (typeof value === "string") {
  if (value.trim() === "") return null;
  const parsed = Number(value);
  return Number.isNaN(parsed) ? null : parsed;
}
```

lib/survey-features/expression-formatting/formatters.ts
```diff
@@ -0,0 +1,161 @@
...
+ * @param params - The parameters for the function.
+ *  [0] - The date to format.
+ *  [1] - The style to format to. Can be "full", "long", "medium", or "short".
+ *  [2] - The locale to format to.
+ * @returns The formatted date
+ */
+function formatDateExpressionFunc(params: unknown[]): string {
+  if (!params || params.length < 1) return EMPTY_STRING;
+
+  const dataStyleValue = getStringParam(params, 1);
```
orilux: nit: `data` instead of `date`

lib/survey-features/expression-formatting/formatters.ts
```diff
@@ -0,0 +1,161 @@
...
+  const locale = getStringParam(params, 2);
+
+  const normalizedDataStyle = dataStyleValue
```
orilux: nit: `data` instead of `date`

lib/survey-features/expression-formatting/bootstrap.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  const factory = FunctionFactory.Instance;
+  for (const { name, func } of expressionFormattingRegistry) {
+    factory.register(name, func, IS_ASYNC, USE_CACHE);
+  }
+
+  areFunctionsRegistered = true;
+}
+
+registerExpressionFormatting();
```
orilux: `registerExpressionFormatting()` is getting executed globally upon import. Doesn't this bypass the new static extension loading lifecycle that was introduced in `use-extension-loader.tsx`, which relies on explicitly invoking the bootstrap method from the registry definition. This might cause SSR hydration issues or premature evaluation before SurveyJS is fully ready.

Recommendation: Remove the top-level `registerExpressionFormatting();` execution at the end of the file.

orilux (changes requested):

lib/survey-features/expression-formatting/bootstrap.ts
PeterMarinov: Good catch. This was left over from the previous implementation of deferred loading with default module import.

lib/survey-features/expression-formatting/formatters.ts
PeterMarinov: mr Data is my obsession

lib/utils/formatters.ts
PeterMarinov: Didn't know that. Resolved and added test

lib/survey-extensions/ui/use-extension-loader.tsx
```diff
@@ -66,29 +123,46 @@ export function useExtensionLoader({
...
+    );
+  }, [allExtensions, normalizedExtensionIds]);
 
   useEffect(() => {
     mountedRef.current = true;
 
-    // Guard to avoid unnecessary loading
-    if (extensionsToLoad.length === 0 || initializedKeys.has(extensionIdsKey)) {
+    // Initialize static extensions synchronously.
+    initializeStaticExtensions(extensionsToLoad);
```
orilux: ### Timing Issue with `setValueExpression`

I noticed that `initializeStaticExtensions(extensionsToLoad)` needs to be executed synchronously outside of the `useEffect` block in `use-extension-loader.tsx`. To see the effects of this, please add the following question to your existing test survey and notice that the value is not getting set when a country is selected:

```
{
          "type": "text",
          "name": "question2",
          "setValueIf": "{country} notempty",
          "setValueExpression": "format({taxRate}, 'percent')"
        }
```

**The Reason:**
When `SurveyModel` is instantiated in the child component (`SurveyComponent`), it immediately parses the JSON schema. During this initial pass, it compiles properties like `setValueExpression` into an Abstract Syntax Tree. 

Because React's `useEffect` runs *after* the render phase (and child effects run before parent effects), SurveyJS attempts to compile the `setValueExpression` *before* the parent component has registered the `formatCurrency` function. 

While standard `expression` questions survive this because they continuously re-evaluate dynamically, `setValueExpression` fails silently during the initial build. Moving the static extension initialization outside of the `useEffect` guarantees the custom functions are registered synchronously and are ready before SurveyJS needs them.

orilux (approved):

PR #543, opened 2026-04-14
fix: Made the "build-embed" script cross-platform

orilux:
## Description
Fixed the "Unsupported URL scheme" error when starting Endatix Hub on Windows

## Related Issues
#542 

## Type of Change
Check all that apply.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Dependency update
- [ ] Refactoring (non-breaking change which improves code quality)
- [ ] Other (please describe)

## Checklist
- My code follows the style guidelines of this project
- I have performed a self-review of my own code
- I have commented my code, particularly in hard-to-understand areas
- I have made corresponding changes to the documentation
- My changes generate no new warnings
- I have added tests that prove my fix is effective or that my feature works
- New and existing tests pass locally with my changes
- Any dependent changes have been merged and published in downstream modules

> [!NOTE]
> - [x] I confirm that my Pull Request follows all the checklist requirements above

## Screenshots
If applicable, add screenshots to help explain your changes.

PR #545, opened 2026-04-14
feat: Add `anyAnswered` and `anyAnsweredByPrefix` logic expression functions

orilux:
## Description
This PR introduces two custom SurveyJS expression functions - `anyAnswered` and `anyAnsweredByPrefix` - designed to replace massive, CPU-blocking chains of `notempty` checks in complex forms. 

During implementation, I identified a critical performance bottlenecks related to how SurveyJS handles reactivity and state cloning. The requirements and syntax for these functions were updated to ensure they run efficiently on low-powered mobile devices without freezing the main UI thread.

## Technical Context
The initial approach allowed form builders to pass question names as strings (e.g., `anyAnswered('q1', 'q2')`). However, this introduced two architectural flaws:
1. **AST Bypass:** Passing strings hides dependencies from the SurveyJS `ExpressionRunner`. Without an Abstract Syntax Tree (AST) dependency graph, SurveyJS forces a global re-evaluation of the expression on *every single keystroke* across the entire form.
2. **The `survey.data` Penalty:** The original functions evaluated state by calling `context.survey.data`. In SurveyJS, `.data` is a computed getter that recursively traverses all elements to generate a deep-cloned JSON payload. Calling this globally on every value change severely blocks the main thread.

### 1. `anyAnswered` (AST-Compliant)
We changed the function signature to evaluate literal values rather than resolving string names. By requiring form builders to use the `{questionName}` syntax, the `ExpressionRunner` successfully registers the exact dependencies and only fires evaluations when those specific questions change. 

**JSON Usage Example:**
```json
{
  "type": "comment",
  "name": "question2",
  "visibleIf": "anyAnswered({qAB101_C103_D_E4_F5}, {qAB101_C103_D_E4_F7}, {qAB101_Z001_D_E4_F17})"
}
```

### 2. `anyAnsweredByPrefix` (Memory-Reference Optimization)
Because dynamic prefix lookups cannot be parsed into an AST at initialization, this function inherently runs globally on every change. To make this safe for mobile CPUs, we dropped the `survey.data` deep-clone entirely. 

The function now calls `survey.getAllQuestions()` and iterates through the direct memory references of the internal question objects, checking their state via the native, highly optimized `.isEmpty()` method.

**JSON Usage Example:**
```json
{
  "type": "comment",
  "name": "question1",
  "visibleIf": "anyAnsweredByPrefix('qAB101_C103_D_E4')"
}
```

## Acceptance Criteria Met
- [x] Replaces long `notempty or notempty` strings.
- [x] `anyAnswered` triggers localized, reactive updates without global state cloning.
- [x] `anyAnsweredByPrefix` safely evaluates without freezing the UI thread.
- [x] Automatically handles complex SurveyJS types (File arrays, Matrices) via `Helpers.isValueEmpty(param)` and `q.isEmpty()`.

## Related Issues
#537 

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [ ] Other (please describe)

## Checklist

> [!NOTE]

### DIFF
--- a/lib/survey-features/any-answered/functions.ts
+++ b/lib/survey-features/any-answered/functions.ts
@@ -1,71 +1,38 @@
-import { SurveyModel } from "survey-core";
+import { Helpers, SurveyModel } from "survey-core";
 
-interface FunctionContext {
+interface ExpressionRunnerContext {
   survey?: SurveyModel;
 }
 
-function isSurveyValueEmpty(value: unknown): boolean {
-  if (value === null || value === undefined) {
-    return true;
-  }
-
-  if (typeof value === "string") {
-    return value.trim().length === 0;
-  }
-
-  if (Array.isArray(value)) {
-    return value.length === 0;
-  }
-
-  return false;
-}
-
-function hasAnsweredValue(
-  questionName: string,
-  values: Record<string, unknown>,
-): boolean {
-  return !isSurveyValueEmpty(values[questionName]);
-}
-
-function getValues(properties: unknown): Record<string, unknown> {
-  if (!properties || typeof properties !== "object") {
-    return {};
-  }
-
-  return properties as Record<string, unknown>;
-}
-
 export function anyAnsweredFunction(
-  this: FunctionContext,
   params: unknown[],
-  properties: unknown,
 ): boolean {
-  const context = this as FunctionContext | undefined;
-  const values = context?.survey?.data ?? getValues(properties);
-  const questionNames = params.filter(
-    (param): param is string => typeof param === "string" && param.length > 0,
-  );
-
-  if (questionNames.length === 0) {
+  if (!Array.isArray(params) || params.length === 0) {
     return false;
   }
 
-  return questionNames.some((questionName) => hasAnsweredValue(questionName, values));
+  return params.some((param) => !Helpers.isValueEmpty(param));
 }
 
 export function anyAnsweredByPrefixFunction(
-  this: FunctionContext,
+  this: ExpressionRunnerContext,
   params: unknown[],
-  properties: unknown,
 ): boolean {
-  const context = this as FunctionContext | undefined;
-  const values = context?.survey?.data ?? getValues(properties);
-  const [prefixParam] = params;
-  if (typeof prefixParam !== "string" || prefixParam.length === 0) {
+  const survey = this?.survey;
+  const prefix = params?.[0];
+
+  if (!survey || typeof prefix !== "string" || prefix.length === 0) {
     return false;
   }
 
-  return Object.keys(values)
-    .filter((questionName) => questionName.startsWith(prefixParam))
-    .some((questionName) => hasAnsweredValue(questionName, values));
+  const questions = survey.getAllQuestions();
+
+  for (let i = 0; i < questions.length; i++) {
+    const question = questions[i];
+    if (question.name.startsWith(prefix) && !question.isEmpty()) {
+      return true;
+    }
+  }
+
+  return false;
 }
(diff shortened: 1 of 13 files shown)


scripts/build-embed.mjs
PeterMarinov, line 9: This file was merged with https://github.com/endatix/endatix-hub/pull/543. Looks like you need to merge main first to avoid conflicts with the build-embed file

lib/survey-features/any-answered/infrastructure/registry.ts
```diff
@@ -0,0 +1,35 @@
...
+    anyAnsweredFunction,
+    false,
+    false,
+  );
+
+  FunctionFactory.Instance.register(
+    ANY_ANSWERED_BY_PREFIX_FUNCTION_NAME,
+    anyAnsweredByPrefixFunction,
+    false,
+    false,
```
PeterMarinov, line 31: I recommend you try with `useCache : true` (for the last param) here and see if it will work. In theory this will further speed up function compute time and CPU given the goal of this optimisation. Might not work OOB, but worth the shot - https://github.com/surveyjs/survey-library/pull/10884/changes

lib/survey-features/any-answered/functions.ts
```diff
@@ -0,0 +1,38 @@
...
+  params: unknown[],
+): boolean {
+  const survey = this?.survey;
+  const prefix = params?.[0];
+
+  if (!survey || typeof prefix !== "string" || prefix.length === 0) {
+    return false;
+  }
+
+  const questions = survey.getAllQuestions();
```
PeterMarinov: Is here the intent to all questions or visible only? Please use the method arguments so it's clear for the future and also to lock this against future API changes

  public getAllQuestions(
    visibleOnly: boolean = false,
    includeDesignTime: boolean = false,
    includeNested: boolean = false
  )

PeterMarinov (changes requested):
Request change for fix the merge conflict with build-embed.js and review the others for the script

lib/survey-features/any-answered/infrastructure/registry.ts
orilux: After some research I decided against enabling on either of the new functions:

anyAnswered uses the {questionName} bracket syntax - SurveyJS already uses its abstract syntax tree to track dependencies. It knows exactly when to re-evaluate without needing function-level cache. Enabling it would likely add unnecessary memory and comparison overhead to a function that is already highly optimized.

anyAnsweredByPrefix exits as soon as it finds a match. SurveyJS's cache would only track the questions checked before the exit. If a user changes a question further down the form, the cache might not invalidate correctly, leading to UI bugs.

lib/survey-features/any-answered/functions.ts
orilux: Done

scripts/build-embed.mjs
orilux: Done

lib/survey-features/any-answered/infrastructure/registry.ts
PeterMarinov: If you don't measure or haven't tried, then this is just an opinion, so I can't accept these as facts. Let's leave it as it is and call it "let's not optimize yet until we see how it performs and can measure it or need to further optimize"

Still... few suggestions that I think are good for us to keep in mind:
-> **anyAnswered** -> surveyJS uses HashTable, so it will have const 0(1) algo complexity. Yes, memory might be the cost, but I won't exclude this as an optimization with expected positive gain even if it is incremental
-> **anyAnsweredByPrefix** -> you go over all questions and filter them by name. That is the most expensive operation and the logic is required once per survey lifecycle given the same prefix argument. It's clear that this can be cached with custom HashTable per prefix to return list of question names or let SurveyJS do it's  caching of the function. Just an idea for the future

PeterMarinov (approved):
Approved

PR #676, opened 2026-06-15
fix(helm): Add Helm Chart

igdragoev:
# Add Helm chart and adopt the Endatix release model

Adds a Helm chart for Endatix Hub and wires the repo into 'endatix/release-workflows`, so the image and chart are built, versioned and published automatically — same model as `endatix/endatix`. Previously nothing was published.

## What ships

| Artifact | Canary (push `main`) | Stable / maintenance |
|---|---|---|
| Image | GHCR | GHCR + Docker Hub |
| Helm chart | GHCR (OCI) | GHCR (OCI) |
| `latest` | never moves | stable only |

Chart and image are lockstep: a chart at `X.Y.Z` deploys the image at `X.Y.Z`.

## Changes

- **`scripts/release-prepare.sh` / `release-publish.sh`** (new) — the consumer
  contract: build image + package chart, then push with the canary/stable split.
- **`release-start.yml`, `release-artifacts.yml`, `release-hotfix.yml`** (new) —
  thin callers of the shared workflows.
- **`build-ci.yml`** — triggers move from `push: "**"` to PRs + `main`/`hotfix/**`;
  adds concurrency, read-only default permissions, and a `pipeline` job gated on
  `needs: [node-build]`. Test job unchanged.
- **`Dockerfile`** — runs as non-root `node` (uid 1000) via `COPY --chown` +
  `USER node`, required for the chart's `runAsNonRoot`.
- **Helm chart** — `_helpers.tpl` adds a release-qualified name and
  release-specific selectors so two releases can coexist in one namespace;
  pod/container security context (non-root, seccomp `RuntimeDefault`, drop all
  caps, read-only rootfs + `/tmp` emptyDir).

## Note for reviewers

Feature-branch pushes no longer trigger CI on their own — builds run on PRs and
on `main`/`hotfix/**`. Update branch protection if it referenced the old
trigger. Image is amd64-only: `sharp`/`esbuild` and `next build` can't be
cross-compiled, so arm64 would mean QEMU on every PR.

### DIFF
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -1,9 +1,38 @@
-name: Build & Test
+name: Build, Test & Release
+
+# Thin consumer of endatix/release-workflows (see its README):
+#   • PR to main → tests (below) + build validation in the shared pipeline
+#     (scripts/release-prepare.sh @ 0.0.0-ci — builds the image and packages
+#     the chart, publishes nothing)
+#   • push main  → tests, then canary vX.Y.Z-canary.N: semantic-release builds
+#     once at the real version and publishes the image + Helm chart to GHCR
+#     (canaries NEVER go to Docker Hub — it is public and immutable).
+#   • stable releases → MANUAL, human-reviewed: run "Start Stable Release"
+#     (release-start.yml) → edit the draft notes → click "Publish release" →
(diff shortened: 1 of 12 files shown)


PeterMarinov (approved):
Approved

PR #718, opened 2026-07-07
feat(h717): Enable lazy loading of data list properties for the property grid

### LINKED ISSUE #717: Property-grid lazy loading for data-list powered priority pickers (carry-forward + loops)
### Goal

Creator property-grid tagboxes for **Priority items** (`advancedCarryForwardPriorityItems`, `priorityItems`) only show static `question.choices`. When a source question uses a data list (400+ items), the dropdown is empty or incomplete at design time.

Centralize lazy-load wiring in the **data-lists** slice and register domain providers from carry-forward and question-loops so both features reuse the same `onChoicesLazyLoad` / `onGetChoiceDisplayValue` pattern already used for `edxDataListId`.

### Implementation Notes

- Extract JWT + API calls from `data-lists/infrastructure/survey-bindings.ts` into a shared module.
- Add a **property-grid lazy choice provider registry** in data-lists; extend survey bindings to dispatch by property name.
- Enable lazy load via `creator.onPropertyGridSurveyCreated` (not `options.element` on `onSurveyInstanceCreated`).
- Carry-forward: provider for `advancedCarryForwardPriorityItems`; static-only `choicesfunc` when any source has `edxDataListId`.
- Question-loops: provider for `priorityItems`; same pattern.
- v1 paging: static prefix + parallel search per data-list source; merge + dedupe by value.
- Requires `data-lists-runtime` extension (form editor JWT). Template editor exclusion unchanged.
- Plan: `.cursor/plans/property_grid_lazy_choices_e5a842b1.plan.md`

PeterMarinov:
# feat(h717): Enable lazy loading of data list properties for the property grid

## Description
- Adds support for lazy loading in the Survey Creator property grid, so that we can enable lazy loading of data lists in data list dependent properties like priorityItems for loops and carry-forward where the source questions use data lists
- Adds priority resolution when a single field uses static + dynamic data sources e.g. static list + one or more data lists + search in all data lists altogether
- Refactor data lists vertical slices to improve DRY and readability

## Related Issues
- closes #717 

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [ ] Other (please describe)

## Checklist

> [!NOTE]

## Screenshots
image

### DIFF
--- a/lib/survey-features/data-lists/infrastructure/property-grid-lazy-choice-registry.ts
+++ b/lib/survey-features/data-lists/infrastructure/property-grid-lazy-choice-registry.ts
@@ -1,4 +1,5 @@
 import type { ExtensionRuntimeDeps } from "@/lib/survey-extensions/types";
+import type { SurveyCreatorModel } from "survey-creator-core";
 import type { Question } from "survey-core";
 import { DEFAULT_CHOICES_LAZY_LOAD_PAGE_SIZE } from "../constants";
 import type {
@@ -37,14 +38,10 @@ export function clearPropertyGridLazyChoiceProvidersForTests(): void {
   providers.clear();
 }
 
-export function setupPropertyGridLazyChoices(
+export function refreshPropertyGridLazyChoices(
   ctx: PropertyGridLazyChoiceContext,
 ): void {
   for (const provider of providers.values()) {
-    if (!provider.shouldEnable(ctx)) {
-      continue;
-    }
-
     const editor = ctx.propertyGridSurvey.getQuestionByName(
       provider.propertyName,
     );
@@ -58,10 +55,34 @@ export function setupPropertyGridLazyChoices(
       choices?: unknown[];
     };
 
+    if (!provider.shouldEnable(ctx)) {
+      tagbox.choicesLazyLoadEnabled = false;
+      continue;
+    }
+
     tagbox.choicesLazyLoadEnabled = true;
     tagbox.choicesLazyLoadPageSize = DEFAULT_CHOICES_LAZY_LOAD_PAGE_SIZE;
     tagbox.choices = provider.getStaticChoices?.(ctx, "") ?? [];
   }
+}
+
+export function refreshPropertyGridLazyChoicesForCreator(
+  creator: SurveyCreatorModel,
+): void {
+  const designerSurvey = creator.survey;
+  const propertyGridSurvey = creator.propertyGrid;
+  const editingObj =
+    creator.selectedElement ?? propertyGridSurvey?.editingObj ?? null;
+
+  if (!designerSurvey || !propertyGridSurvey || editingObj == null) {
+    return;
+  }
+
+  refreshPropertyGridLazyChoices({
+    designerSurvey,
+    propertyGridSurvey,
+    editingObj,
+  });
 }
 
 export async function dispatchPropertyGridChoicesLazyLoad(
--- a/lib/survey-features/data-lists/infrastructure/property-grid-lazy-choice-registry.ts
+++ b/lib/survey-features/data-lists/infrastructure/property-grid-lazy-choice-registry.ts
@@ -1,0 +1,96 @@
+import type { ExtensionRuntimeDeps } from "@/lib/survey-extensions/types";
+import type { Question } from "survey-core";
+import { DEFAULT_CHOICES_LAZY_LOAD_PAGE_SIZE } from "../constants";
+import type {
+  PropertyGridLazyChoiceContext,
+  PropertyGridLazyChoicePageParams,
+  PropertyGridLazyChoiceProvider,
+} from "../types";
+
+export type {
+  PropertyGridLazyChoiceContext,
+  PropertyGridLazyChoicePageParams,
+  PropertyGridLazyChoiceProvider,
+} from "../types";
+
+const providers = new Map<string, PropertyGridLazyChoiceProvider>();
+
+export function registerPropertyGridLazyChoiceProvider(
+  provider: PropertyGridLazyChoiceProvider,
+): void {
+  providers.set(provider.propertyName, provider);
+}
+
+export function unregisterPropertyGridLazyChoiceProvider(
+  propertyName: string,
+): void {
+  providers.delete(propertyName);
+}
+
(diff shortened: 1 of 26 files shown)


lib/survey-features/data-lists/use-cases/load-choices-in-creator.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  if (params.skip < staticItems.length) {
+    const staticPage = staticItems.slice(
+      params.skip,
+      params.skip + params.take,
+    );
+    const apiTake = params.take - staticPage.length;
+
+    if (apiTake <= 0) {
+      return { items: staticPage, total: staticItems.length };
```
orilux: `total` nay get under-reported when a page is all static items. When `apiTake <= 0` the early return uses `total: staticItems.length` only. If `dataListSources` is non-empty, the real total is `staticItems.length + <data-list total>`. SurveyJS uses `total` for pagination, so it will stop after the static segment and never load data-list choices.

Example: 10 static + 400 data-list items, `take=25`, `skip=0` → returns 10 items with `total: 10` instead of `410`.

**Fix:** Compute the merged data-list total (even when not fetching items on this page) and return `total: staticItems.length + dataListTotal`. Add a test for the `apiTake <= 0` case.

lib/survey-features/data-lists/use-cases/load-choices-in-creator.ts
```diff
@@ -0,0 +1,171 @@
...
+    formatLabel,
+  );
+
+  return {
+    items: apiItems.items,
+    total: staticItems.length + apiItems.total,
+  };
+}
+
+async function fetchMergedDataListChoices(
```
orilux, line 79: Merged `total` may overcount when sources share values. In `fetchMergedDataListChoices`, `total` is the sum of each source’s API `total`, but `appendUniqueChoices` deduplicates by value across sources. If two data lists overlap, `total` can exceed the number of distinct choices SurveyJS can actually page through.

Example: two lists of 100 with 20 shared values → `total: 200`, but only ~180 unique items exist.

Either document this as a known limitation, or adjust `total` to reflect deduped count (harder without a full scan). Worth a test asserting expected pagination behavior when sources overlap.

lib/survey-features/data-lists/use-cases/load-choices-in-creator.ts
```diff
@@ -0,0 +1,171 @@
...
+      skip: remainingSkip,
+      take: remainingTake,
+    });
+
+    const sourceTotal = resolveSourceTotal(response);
+    if (sourceTotal !== null) {
+      total += sourceTotal;
+    }
+
+    if (!response.success) {
```
orilux, line 107: Failed source doesn’t advance skip for later sources. When `searchDataListChoices` fails, `sourceTotal` is `null` so `remainingSkip` is not reduced past that source. Pagination still works at `skip=0` (covered by test), but deep pages can break: e.g. source A (100 items) fails, source B succeeds, client requests `skip=120` — B is queried with `skip=120` instead of `skip=20`.

Consider skipping past failed sources when size is unknown (best-effort), surfacing partial failure in the UI, or adding a test for `skip > firstSourceTotal` with first source failing.

lib/survey-features/data-lists/infrastructure/property-grid-lazy-choice-registry.ts
```diff
@@ -0,0 +1,117 @@
...
+      continue;
+    }
+
+    const tagbox = editor as Question & {
+      choicesLazyLoadEnabled?: boolean;
+      choicesLazyLoadPageSize?: number;
+      choices?: unknown[];
+    };
+
+    if (!provider.shouldEnable(ctx)) {
```
orilux, line 58: Disabling lazy load doesn’t restore static choices. When `shouldEnable(ctx)` is false, we set `choicesLazyLoadEnabled = false` and `continue` without updating `tagbox.choices`. While lazy load was active, the Serializer `choices` callback returned `[]`, so the editor may stay empty when the user switches to static-only sources (e.g. changes `loopSource` / `edxCarryForwardSources` away from data-list questions).

**Fix:** When disabling, repopulate from static sources (e.g. `getStaticChoicesFromSources` / provider helper) or trigger a property-grid refresh that re-invokes the Serializer `choices` callback. Please verify manually: open priority items with a data-list source, then switch to static-only sources while the property grid stays open.

orilux (changes requested):

lib/survey-features/data-lists/use-cases/load-choices-in-creator.ts
PeterMarinov: The feedback is valid, with one nuance - apiTake cannot be negative. It is params.take - staticPage.length, and staticPage is a slice of length at most params.take, so apiTake >= 0 always. The old apiTake <= 0 guard was really the apiTake === 0 case.

Fixed

lib/survey-features/data-lists/use-cases/load-choices-in-creator.ts
PeterMarinov: Feedback is valid: total sums per-source API counts while items are deduped. Documenting this and adding tests that lock in the current pagination behavior as known limitation

lib/survey-features/data-lists/use-cases/load-choices-in-creator.ts
PeterMarinov: Feedback is valid: on failure we don't advance remainingSkip past the failed source. Fix applied to probe total on failure when remainingSkip > 0, then adding a deep-skip test.

lib/survey-features/data-lists/infrastructure/property-grid-lazy-choice-registry.ts
PeterMarinov: Valid. With lazy load on, choicesfunc returns [], so the tagbox stays empty. Turning lazy load off without repopulating choices left it empty after switching to static-only sources.

Fix: When shouldEnable is false, set choices from provider.getStaticChoices(ctx, "") (same helper used when enabling — static-only segment from loop/carry-forward providers).

orilux (approved):

PR #812, opened 2026-07-28
feat(h727): Drag to categorize question type

### LINKED ISSUE #727: "Drag to Categorize" question type
## Summary

Add a new **Drag to Categorize** question type to the SurveyJS platform in hub. Respondents drag items (text labels or images) from a source pool into one or more named category zones - without implying order or ranking.

## Background

The existing ranking question (`selectToRankEnabled`) provides a single anonymous bucket. Surveys that need respondents to sort items into *named, semantically distinct* categories (e.g. "Agree / Neutral / Disagree", "Must have / Nice to have / Not needed") have no native option today.

## Acceptance Criteria

- [ ] Respondents can drag items from a pool into predefined named drop zones
- [ ] Items render as text labels or uploaded images (stored in blob storage, not base64)
- [ ] Per-item flag: `allowMultipleZones` — clones the item into multiple zones; source stays in pool
- [ ] Per-zone constraints: `minItems` / `maxItems` — validated before page advance
- [ ] Global flag: `requireAllItems` — all items must be placed before advancing
- [ ] Question type appears in the Creator toolbox under the **Choice** group
- [ ] Scripters can configure items, zones, and rules in the visual property grid
- [ ] JSON editor and property grid stay in sync bidirectionally
- [ ] Stored value is `Record<string, string[]>` mapping zone value → placed item values
- [ ] Native SurveyJS conditional logic works: `{q1.zone_a} contains 'item_1'`
- [ ] Works on touch devices (no page scroll during drag)
- [ ] Images scale responsively across viewport sizes
- [ ] Submission viewer renders items statically in their zones (no drag handles)
- [ ] The extension bundle is only loaded for forms that contain a `dragcategorize` question

Key technical decisions:
- **No new dependencies.** Builds on SurveyJS's own drag engine (`DragDropRankingSelectToRank`), requiring only 4 method overrides to support N named zones instead of 2.
- Follows the `ExtensionModule` framework (`lib/survey-extensions/`). Registered in `core-registry.ts` as a `dynamic` extension. No changes to `form-editor.tsx`.
- New feature lives in `lib/survey-features/drag-categorize/` per the add-survey-feature skill.

orilux:
## Description

- Introduced a drag-to-categorize question type
- Allows for both text and images to be categorized in different buckets
- Supports advanced carry forward
- Optional captions for images
- Responsive
- Renders images in submission details and PDF export

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [ ] Other (please describe)

## Checklist

> [!NOTE]

## Screenshots
image

## Summary by CodeRabbit

* **New Features**
  * Added a drag-and-categorize question type with configurable zones, item limits, labels, images, multi-zone placement, and carry-forward support.
  * Added interactive authoring, runtime responses, submission answer views, and PDF export rendering.
  * Added support for preserving and displaying answers when choices or zones change.
  * Improved copying of choice images between question types.
* **Bug Fixes**
  * Protected uploaded images and authorization references in drag-and-categorize questions.
  * Ensured image-only answers display correctly without exposing internal values.
* **Tests**
  * Added comprehensive coverage for authoring, dragging, validation, carry-forward, submissions, and PDF output.


lib/questions/drag-categorize/__tests__/drag-categorize.validation.test.ts
```diff
@@ -0,0 +1,109 @@
...
+      zones,
+    };
+
+    // Act
+    const errors = validateZoneConstraints(source);
+
+    // Assert
+    expect(errors).toHaveLength(1);
+    expect(errors[0].getText()).toContain('"Zone B" allows at most 2');
+  });
```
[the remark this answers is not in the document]
orilux: Addressed

lib/questions/drag-categorize/__tests__/drag-categorize.validation.test.ts
```diff
@@ -0,0 +1,109 @@
...
+      items,
+      zones: [{ value: "zone_x", minItems: 1 }],
+    };
+
+    // Act
+    const errors = validateZoneConstraints(source);
+
+    // Assert
+    expect(errors[0].getText()).toContain('"zone_x"');
+  });
```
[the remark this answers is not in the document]
orilux: Addressed

lib/questions/drag-categorize/constants.ts
```diff
@@ -0,0 +1,25 @@
...
+export const DRAG_CATEGORIZE_TYPE = "dragcategorize";
+
+/** Human-readable name shown in the toolbox and the question type dropdown. */
+export const DRAG_CATEGORIZE_DISPLAY_NAME = "Drag to Categorize";
+
+export const DRAG_CATEGORIZE_ITEM_CLASS = "dragcategorizeitem";
+export const DRAG_CATEGORIZE_ZONE_CLASS = "dragcategorizezone";
+
+/** Virtual zone id for the unplaced-items pool (never serialized into the value). */
+export const POOL_ZONE_ID = "__pool__";
```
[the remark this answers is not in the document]
orilux: Addressed with a dev warning

lib/questions/drag-categorize/drag-categorize.component.tsx
```diff
@@ -0,0 +1,228 @@
...
+          key="newitem"
+          className="sv-categorize__item sv-categorize__item--ghost"
+          onClick={() => question.addItemFromDesigner()}
+        >
+          <div className="sv-categorize__item-content">
+            <span className="sv-categorize__item-text">+ Add an item</span>
+          </div>
+        </div>
+      );
+    }
```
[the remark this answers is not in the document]
orilux: Addressed

lib/questions/drag-categorize/drag-categorize.drag-controller.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  protected findDropTargetNodeByDragOverNode(
+    dragOverNode: HTMLElement,
+  ): HTMLElement {
+    if (dragOverNode.dataset?.categorize) return dragOverNode;
+    // The engine treats a null return as "ban drop here".
+    return dragOverNode.closest<HTMLElement>(
+      "[data-categorize]",
+    ) as HTMLElement;
+  }
```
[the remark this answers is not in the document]
orilux: Rejected - this would break the base-class override contract

lib/questions/drag-categorize/utils/zone-helpers.ts
```diff
@@ -0,0 +1,226 @@
...
+ *
+ * A non-clone item is cleared from every zone it currently occupies rather
+ * than from the zone the drag reported starting in. The placement is the
+ * source of truth; the drag origin is transient state that says nothing about
+ * where the item actually sits. This keeps "at most one zone" true even when
+ * the item started out in several — which happens when allowMultipleZones is
+ * turned off after the item was already placed — so the next drag repairs it.
+ *
+ * Clone items are the opposite: each copy is independent, so only the zone the
+ * copy was dragged out of may lose it.
```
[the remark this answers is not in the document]
orilux: Addressed

lib/questions/drag-categorize/utils/zone-helpers.ts
```diff
@@ -0,0 +1,226 @@
...
+      return true;
+    });
+    if (kept.length !== items.length) {
+      next[zoneId] = kept;
+      changed = true;
+    }
+  }
+
+  return { placement: next, removed, changed };
+}
```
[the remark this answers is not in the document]
orilux: Addressed

lib/questions/drag-categorize/__tests__/drag-categorize.validation.test.ts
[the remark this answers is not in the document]
orilux: Sorry, I mistyped "addressed" initially, but corrected it afterwards. Didn't mean to ask you to address it yourself.

lib/questions/drag-categorize/drag-categorize.component.tsx
```diff
@@ -0,0 +1,228 @@
...
+          {item.imageUrl ? (
+            <>
+              <img
+                src={item.imageUrl}
+                // Empty alt when no label was authored: the fallback would be
+                // the item's value ("item1"), which is meaningless to read out.
+                alt={getDisplayLabel(item)}
+                className="sv-categorize__item-img"
+                draggable={false}
+              />
```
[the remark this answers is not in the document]
orilux: Addressed

lib/questions/drag-categorize/__tests__/drag-categorize.validation.test.ts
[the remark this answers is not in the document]
orilux: Applied, with one adjustment. The placed `item_3` is now a defined item.

Couldn't use only the shared collection literally - it has two items and `zone_b`
has `maxItems: 2`, so exceeding the capacity needs a third distinct value. Growing
the shared fixture would break the two `requireAllItems` tests that depend on
exactly what's placed, and repeating a value would test duplicate handling instead
of capacity. So the third item is defined locally in this test.

Skipped the follow-up suggestion. Min/max count raw `.length` and never filter
against `items`, so unknown entries counting toward capacity is incidental rather
than intended - a named test would lock in behaviour that's arguably a bug (stale
data inflating a capacity error). It's also unreachable in practice, since
`clearIncorrectValues` strips unknown items before validation runs.

.agents/skills/add-survey-feature/SKILL.md
igdragoev: This points to file that is not in the repo

features/public-form/application/submission-queue/use-submission-queue.hook.ts
```diff
@@ -1,28 +1,56 @@
...
-    clearQueue,

-    waitForInFlightPartial,

-  };

-}

+import { useCallback } from "react";
+
```
igdragoev, line 2: something odd with the formatting of this file. Too many empty lines

lib/questions/drag-categorize/drag-categorize.extension.ts
```diff
@@ -0,0 +1,23 @@
+import type { ExtensionModule } from "@/lib/survey-extensions/types";
+import { registerDragCategorizeQuestion } from "./drag-categorize.component";
+import { bindDragCategorizeToCreator } from "./drag-categorize.creator";
```
igdragoev: **`survey-creator-core` gets bundled into the chunk the public respondent runner downloads.**

This dynamically-loaded extension module statically imports the Creator bindings.

A respondent opens a public form containing a `dragcategorize` question with `ENDATIX_ENABLE_EXTENSIONS=true`. `survey-js-wrapper.tsx` → `useSurveyExtensions({formJson})` → `shouldLoad: analyzer.usesQuestionType('dragcategorize')` → `import('.../drag-categorize.extension')`. That module's top-level `import { bindDragCategorizeToCreator } from './drag-categorize.creator'` pulls `drag-categorize.creator.ts`, which does `import { SvgRegistry } from 'survey-core'` and `import { getLocaleStrings } from 'survey-creator-core'`. Webpack therefore emits the whole SurveyJS Creator into a chunk fetched on the respondent path, where it is never used — the runner already registers the question synchronously in `use-survey-model.hook.ts`.

This is the exact regression the PR's own comments in `use-survey-model.hook.ts:6-8` and `submission-survey.tsx:17-19` deep-import to avoid. No other core extension imports `survey-creator-core`.

**Fix:** make `onCreatorReady` do `const { bindDragCategorizeToCreator } = await import('./drag-categorize.creator')`, or split the Creator adapter into its own extension module.

lib/questions/drag-categorize/utils/zone-helpers.ts
```diff
@@ -0,0 +1,281 @@
...
+/**
+ * Normalizes an arbitrary question value into a safe placement record.
+ * Drops non-array zone entries and non-scalar item values.
+ */
+export function parsePlacement(raw: unknown): DragCategorizePlacement {
+  if (!raw || typeof raw !== "object" || Array.isArray(raw)) {
+    return {};
+  }
+
+  const placement: DragCategorizePlacement = {};
```
igdragoev: **Prototype pollution: `placement[zoneId] ?? []` can return an inherited `Object.prototype` member instead of an array.**

`parsePlacement` builds the placement map as a plain `{}`.

A form author names a zone `constructor` (or `toString`, `valueOf`, `hasOwnProperty`). The stored value has no own key for it, so `placement['constructor']` returns the `Object` function — non-nullish, so `?? []` does not fire. Consequences:

- `DragCategorizeQuestion.getZoneItems` (`drag-categorize.model.ts:181-186`) does `placed.map(...)` on a function → `TypeError: placed.map is not a function`, crashing the runner render.
- The same expression in `resolveAnswerZones` (`utils/answer-display.ts:103`) crashes the submission-details viewer and the PDF export via `resolveItems`.
- `validateZoneConstraints` (`drag-categorize.validation.ts:55`) computes `(placement[zone.value] ?? []).length` → `Object.length === 1`, silently mis-evaluating min/max.

Separately, a stored placement keyed `__proto__` is lost: `placement['__proto__'] = [...]` invokes the prototype setter rather than creating an own key.

**Fix:** use `Object.create(null)` for the returned record (or guard every read with `Object.hasOwn`).

lib/questions/drag-categorize/utils/zone-helpers.ts
```diff
@@ -0,0 +1,281 @@
...
+  zoneIds: string[],
+  multiZoneItems: Set<string>,
+): boolean {
+  const seen = new Set<string>();
+  let changed = false;
+
+  for (const zoneId of zoneIds) {
+    const items = next[zoneId];
+    if (!items) continue;
+    const kept = items.filter((itemValue) => {
```
igdragoev, line 247: **Duplicate item values within one zone are never deduped for `allowMultipleZones` items — duplicate React keys and an undeletable phantom chip.**

`enforceSingleZone` short-circuits with `if (multiZoneItems.has(itemValue)) return true;` before the `seen` check, and `parsePlacement` only filters non-scalars. So a stored value of `{zone_a: ["x", "x"]}` for an item with `allowMultipleZones: true` — reachable from legacy/hand-edited submission JSON, an external `survey.data = {...}` write, or an API-imported response — survives reconciliation untouched.

`getZoneItems('zone_a')` then returns the same `ItemValue` twice; `DragCategorizeComponent.renderItem` emits two elements with the identical ``key={`${zoneId}:${item.value}`}`` and `DragCategorizeAnswer` two `<Badge key={item.value}>` — React logs duplicate-key warnings and can reuse the wrong node. Dragging either chip out calls `removeFrom`, which filters *all* matching values, so the second chip vanishes with the first and the count is never repairable one-at-a-time.

**Fix:** dedupe per zone in `parsePlacement`.

lib/questions/drag-categorize/drag-categorize.registry.ts
```diff
@@ -0,0 +1,162 @@
...
+      () => new DragCategorizeQuestion(""),
+      "selectbase",
+    );
+  }
+
+  // Opt into advanced Carry forward now that the class exists. Must stay
+  // after addClass: SurveyJS discards properties added to an unknown class,
+  // and carry forward (a static extension) initializes before this question's
+  // dynamic extension, so it cannot register us from its own side.
+  registerCarryForwardForQuestionType(DRAG_CATEGORIZE_TYPE);
```
igdragoev: **The server-safe model registry unconditionally opts drag-categorize into carry-forward, an asymmetry with the built-in types it copies.**

For built-in types, `CARRY_FORWARD_PROPERTIES` are only added by `registerAdvancedCarryForwardGlobals()`, which runs inside the `carryForwardExtension.onInit` hook and is therefore gated by `ENDATIX_ENABLE_EXTENSIONS`. `registerDragCategorizeModel()` calls `registerCarryForwardForQuestionType(DRAG_CATEGORIZE_TYPE)` on *every* invocation — the public runner (`use-survey-model.hook.ts`), the read-only submission survey, and the server-side PDF pipeline (`prepare-pdf-model.use-case.ts:33`).

On a deployment with extensions off, a form JSON carrying `edxCarryForwardEnabled: true` now makes `isAdvancedCarryForwardEnabled()` return true (via the new `isSupportedCarryForwardQuestionType` path) and `configureNativeChoiceSourceMutualExclusionForType` hides `choicesFromQuestion`/`choicesByUrl`, yet nothing ever runs `syncSingleCarryForwardTarget` — the question renders with an empty item pool and no diagnostic.

It also makes the server-only registry import and mutate an unrelated feature's global state on every PDF export.

lib/questions/drag-categorize/drag-categorize.model.ts
```diff
@@ -0,0 +1,539 @@
...
+
+  // --- Validation and value hygiene ---
+
+  protected onCheckForErrors(
+    errors: SurveyError[],
+    isOnValueChanged: boolean,
+    fireCallback: boolean,
+  ): void {
+    super.onCheckForErrors(errors, isOnValueChanged, fireCallback);
+    if (this.isEmpty() && !this.requireAllItems) return;
```
igdragoev, line 338: **`minItems`/`maxItems` are only enforced once the respondent places at least one item *anywhere*.**

`onCheckForErrors` returns before running any zone validation whenever the question is empty.

Given a non-required question with `zones: [{value:'a', minItems:2}, {value:'b'}]` and `requireAllItems: false`:

- Respondent A places nothing → `this.isEmpty()` is true and `!this.requireAllItems` is true → early return → the page validates and advances with zone A empty.
- Respondent B drags a single item into zone B → the question is no longer empty → `validateZoneConstraints` runs and reports `"a" needs at least 2 item(s).` for a zone they never touched, blocking them.

The same interaction makes a `maxItems` violation loaded from pre-existing submission data invisible until the value is otherwise non-empty.

**Fix:** the gate should be about whether the respondent skipped an optional question, not a blanket bypass of every per-zone constraint.

lib/questions/drag-categorize/drag-categorize.creator.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  // Categorization needs at least MIN_ZONES zones — block deleting below it.
+  creator.onCollectionItemAllowOperations.add((_, options) => {
+    if (
+      options.propertyName !== ZONES_PROPERTY ||
+      options.element?.getType() !== DRAG_CATEGORIZE_TYPE
+    ) {
+      return;
+    }
+    if ((options.collection?.length ?? 0) <= MIN_ZONES) {
```
igdragoev: **A missing `options.collection` is treated as length 0, which forbids the delete.**

`creator.onCollectionItemAllowOperations` can fire for the `zones` property with `options.collection` undefined — the event contract does not guarantee it for every operation kind, and this PR's tests always construct it by hand. `(undefined?.length ?? 0) <= MIN_ZONES` evaluates `0 <= 2` → `options.allowDelete = false`, so a form author with 5 zones cannot delete any of them and gets no explanation.

**Fix:** the fallback should be an early `return` (leave the decision alone) rather than a value that satisfies the block condition.

lib/questions/drag-categorize/drag-categorize.creator.ts
```diff
@@ -0,0 +1,106 @@
...
+ * property-grid help texts). Safe to call repeatedly.
+ *
+ * Per-creator wiring lives in bindDragCategorizeToCreator, which needs the
+ * SurveyCreator instance.
+ */
+export function registerDragCategorizeQuestionUI(): void {
+  registerDragCategorizeQuestion();
+  SvgRegistry.registerIcon(DRAG_CATEGORIZE_TYPE, DRAG_CATEGORIZE_SVG);
+
+  const translations = getLocaleStrings("en");
```
igdragoev, line 20: **Toolbox name and property-grid help are written only into the hard-coded `"en"` locale bucket.**

A designer runs the Creator with a non-English locale (`editorLocalization.currentLocale = 'de'`, or the app sets a locale from user settings). `getLocaleStrings("en")` returns the English bucket only, so `translations.qt["dragcategorize"]` and `translations.pehelp["dragcategorize"]` are never consulted at render time — `getLocalizedToolboxItemName` resolves `qt.dragcategorize` against the active locale, misses, and the toolbox plus the question-type dropdown display the literal string `dragcategorize`. All five `pehelp` descriptions disappear.

**Fix:** either write into `getLocaleStrings(editorLocalization.currentLocale)` as well, or register the strings for the default locale bucket that other locales fall back through.

lib/utils/survey/copy-choice-item.ts
```diff
@@ -1,4 +1,38 @@
...
+/**
+ * Property names choice items use for their image, in read preference order.
+ *
+ * SurveyJS's imagepicker calls it `imageLink` on `imageitemvalue`; the
+ * drag-categorize item calls it `imageUrl`. Neither class declares the other,
+ * so carrying a choice across those types has to translate the name — copying
+ * `imageLink` onto a drag-categorize item stores a property nothing reads, and
+ * the chip renders as text only.
+ */
+const IMAGE_PROPERTIES = ["imageLink", "imageUrl"] as const;
```
igdragoev: **A shared cross-feature utility now hard-codes drag-categorize's `imageUrl` property name.**

`IMAGE_PROPERTIES = ["imageLink", "imageUrl"]` is a fixed two-entry list: `findImageProperty` is Serializer-driven (as its doc claims) but `readItemImage` is not, so any future code-owned item class that names its image something else must edit this shared file again, and the next question type after that adds a third entry.

`lib/utils/survey/__tests__/copy-choice-item.test.ts` additionally now does `import { registerDragCategorizeModel } from "@/lib/questions/drag-categorize/drag-categorize.registry"`, making a `lib/utils` unit test depend on a specific question's registration. Per `AGENTS.md` ("Put reusable cross-feature utilities in `lib/`; keep feature-specific business logic inside the owning feature slice") the property-name knowledge belongs with the question.

**Fix:** derive the source property the same Serializer-driven way `findImageProperty` already does — `Serializer.getProperties(choice.getType())` filtered to `type === "file"` — instead of a hand-maintained list.

lib/questions/drag-categorize/drag-categorize.component.tsx
```diff
@@ -0,0 +1,240 @@
...
+                  zoneId,
+                  event.currentTarget,
+                );
+              }
+            : undefined
+        }
+      >
+        <div
+          className={`sv-ranking-item__content sv-categorize__item-content${
+            item.imageUrl && this.getItemCaption(item)
```
igdragoev: **`renderItem` recomputes the item caption three times per chip per render, once just to pick a CSS class.**

For every chip on every render the component calls `this.getItemCaption(item)` at line 194 (className branch), again at line 210 (visibility branch), and `getDisplayLabel(item)` a third time at line 206 (alt text) — each walking `hasExplicitLabel` → `item.text?.trim()` → `String(item.value).trim()`. During a drag, `hoveredZoneId` changes re-render the whole question, so this runs for the pool plus every zone on each drop-target change.

**Fix:** hoist `const caption = this.getItemCaption(item); const label = getDisplayLabel(item);` to the top of the non-ghost branch and reuse them.

.agents/skills/add-survey-feature/SKILL.md
orilux: Addressed

features/public-form/application/submission-queue/use-submission-queue.hook.ts
orilux: Addressed

lib/questions/drag-categorize/drag-categorize.extension.ts
orilux: Agreed on the fix - I'll move `bindDragCategorizeToCreator` behind an `await import()`
in `onCreatorReady` (or split the Creator adapter into its own module).

One correction to the premise though: "No other core extension imports
`survey-creator-core`" isn't accurate. `carryForwardExtension` and
`dataListsExtension` are both `loading: "static"` in `core-registry.ts`, and both
reach `survey-creator-core` via `infrastructure/creator-help.ts`
(`import { getLocaleStrings }`). `use-survey-extensions.ts` is `"use client"` and
statically imports `coreExtensions`, so the Creator is already in the main
respondent bundle independently of this extension - webpack shares that module with
the dynamic chunk rather than emitting a second copy.

So this is correctness/hygiene rather than a new payload regression on the runner.
Worth doing regardless; and if dropping the Creator from the respondent bundle is the
actual goal, those two static edges are what need fixing.

lib/questions/drag-categorize/utils/zone-helpers.ts
orilux: Addressed

lib/questions/drag-categorize/utils/zone-helpers.ts
orilux: Addressed

lib/questions/drag-categorize/drag-categorize.registry.ts
orilux: Addressed

lib/questions/drag-categorize/drag-categorize.model.ts
orilux: Half agree.

The A/B asymmetry is real, but it's the standard SurveyJS contract for an optional
question: skip it entirely and it validates; engage with it and its constraints
apply. Checkbox behaves the same way with `minSelectedChoices`. The alternatives are
enforcing zone minimums against a respondent who deliberately skipped the question,
or dropping them once it's partially filled. Happy to change it if that's the product
call, but it's a decision rather than a defect.

The second claim doesn't hold: a `maxItems` violation loaded from submission data
can't be hidden by this gate. Exceeding a zone maximum requires items in that zone,
so `this.isEmpty()` is false and `validateZoneConstraints` runs. There's no state
where the value violates `maxItems` and is simultaneously empty.

lib/questions/drag-categorize/drag-categorize.creator.ts
orilux: Taking the fix - an early `return` when `collection` is missing is strictly safer
than a fallback that satisfies the block condition, and costs nothing.

For the record, the reachability isn't demonstrated. `CollectionItemAllowOperationsEvent`
declares `collection: Base[]` (non-optional), and the handler only acts when
`propertyName === "zones"` on a `dragcategorize` element, where `zones` is always
materialised by `createItemValues(ZONES_PROPERTY)` in the constructor. I can't
reproduce the "5 zones and cannot delete any" symptom. Applying it as defensive
hardening rather than a confirmed bug.

lib/questions/drag-categorize/drag-categorize.creator.ts
orilux: Challenging this one - the fallback already works.

`editorLocalization.getString` falls through to the default (`en`) bucket when the
active locale has no entry. Verified against the real translation bundle
(`survey-creator-core.i18n.js`):

    locale=en  qt.checkbox="Checkboxes"      qt.dragcategorize="Drag to Categorize"
    locale=de  qt.checkbox="Auswahl"         qt.dragcategorize="Drag to Categorize"
    locale=fr  qt.checkbox="Cases à cocher"  qt.dragcategorize="Drag to Categorize"
    locale=es  qt.checkbox="Caja"            qt.dragcategorize="Drag to Categorize"

The toolbox never shows the raw `dragcategorize` type name, and the `pehelp` entries
resolve the same way. (`getLocalizedToolboxItemName` also isn't an export of
`survey-creator-core`.)

What is true is milder: the name and help stay English for non-English designers - a
missing translation rather than a broken label, and the same gap every custom
question in this repo has. Worth a ticket for localising custom-question strings, but
not a defect in this PR.

lib/utils/survey/copy-choice-item.ts
orilux: Addressed

lib/questions/drag-categorize/drag-categorize.component.tsx
orilux: Addressed

PR #823, opened 2026-08-03
feat(h819): Auto-select storage and hide the base64 option when a storage provider is enabled

### LINKED ISSUE #819: Hide the SurveyJS Creator option for storing file uploads as text (base64)
# Description

SurveyJS creator provides a checkbox for selecting whether a file upload is going to store data within the submission JSON as a Base64 encoded string or to file storage. This creates confusion among form scripters.

# Expected behavior

- No checkbox is available
- If file storage is enabled automatically route all file uploads to it
- If file storage is not enabled automatically forward all uploads to submission JSON

# Screenshot

Image

orilux:
## Summary

- When a storage provider is enabled, `file` and `signaturepad` questions now always upload to storage instead of embedding content as base64 in the JSON result.
- Hides the `storeDataAsText` property from the Survey Creator property grid in that case, so form authors can't flip it back to text-embedding.
- Applies to both Endatix forms and form templates (both go through the same `useStorageWithCreator` hook).
- No-op when no storage provider is configured - existing behavior (embed as text) is unchanged.

## Implementation

- New `lib/survey-features/storage-only-file-mode/` module:
  - `registry.ts` - hides `storeDataAsText` from the Creator property grid (leaves the SurveyJS default untouched - see note below).
  - `survey-bindings.ts` - forces `storeDataAsText = false` on existing file/signaturepad questions and any added afterwards.
  - `creator-bindings.ts` - binds the above to every Creator survey instance (designer + preview tabs), reusing the existing `bindSurveyToCreatorAreas` helper.
- Wired into `useStorageWithCreator` (`features/asset-storage`), gated by the existing `storageConfig?.isEnabled` check - no changes to `form-editor.tsx` / `form-template-editor.tsx`.

## Note

- Deliberately does **not** change the Serializer's default value for `storeDataAsText` (only hides it). Flipping the default would make an explicit `false` equal to "default" and get dropped from saved JSON - then a respondent session (which never runs this hub-only registry code) would fall back to the built-in default of `true` and silently re-embed files as base64. Covered by a regression test asserting the value round-trips through `toJSON()`.
- Enforcement is Creator-only by design (not applied at runtime/respondent side) - scoped per discussion.

## Related Issues
Closes #819 

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [ ] Other (please describe)

## Checklist

> [!NOTE]

## Screenshots
image

## Summary by CodeRabbit

* **New Features**
  * Added storage-only file handling for file upload and signature-pad questions.
  * New questions and existing questions automatically use external storage instead of embedding data in survey results.
  * Added integration with Survey Creator and survey editing workflows.
  * Added cleanup support to restore standard behavior when storage handling is disabled or removed.

* **Tests**
  * Added coverage for registration, creator and survey bindings, nested questions, cleanup, and default-value preservation.

### DIFF
--- a/lib/survey-features/storage-only-file-mode/constants.ts
+++ b/lib/survey-features/storage-only-file-mode/constants.ts
@@ -7,6 +7,7 @@ export type StorageOnlyFileModeQuestionType =
   (typeof STORAGE_ONLY_FILE_MODE_QUESTION_TYPES)[number];
 
 export const STORE_DATA_AS_TEXT_PROPERTY = "storeDataAsText";
+export const WAIT_FOR_UPLOAD_PROPERTY = "waitForUpload";
 
 export const STORAGE_ONLY_FILE_MODE_CREATOR_BOUND_KEY =
   "__endatixStorageOnlyFileModeCreatorBound";
--- a/lib/survey-features/storage-only-file-mode/constants.ts
+++ b/lib/survey-features/storage-only-file-mode/constants.ts
@@ -1,0 +1,12 @@
+export const STORAGE_ONLY_FILE_MODE_QUESTION_TYPES = [
+  "file",
+  "signaturepad",
+] as const;
+
+export type StorageOnlyFileModeQuestionType =
+  (typeof STORAGE_ONLY_FILE_MODE_QUESTION_TYPES)[number];
+
+export const STORE_DATA_AS_TEXT_PROPERTY = "storeDataAsText";
+
+export const STORAGE_ONLY_FILE_MODE_CREATOR_BOUND_KEY =
+  "__endatixStorageOnlyFileModeCreatorBound";
--- a/lib/survey-features/storage-only-file-mode/infrastructure/registry.ts
+++ b/lib/survey-features/storage-only-file-mode/infrastructure/registry.ts
@@ -2,44 +2,61 @@ import { Serializer } from "survey-core";
 import {
   STORAGE_ONLY_FILE_MODE_QUESTION_TYPES,
   STORE_DATA_AS_TEXT_PROPERTY,
+  WAIT_FOR_UPLOAD_PROPERTY,
 } from "../constants";
+
+// Both are forced by bindStorageOnlyFileModeToSurvey (storeDataAsText: false,
+// waitForUpload: true). Hide both from the grid — leaving either editable
+// lets an author flip it back and silently defeat the enforcement.
+const HIDDEN_PROPERTIES = [
+  STORE_DATA_AS_TEXT_PROPERTY,
+  WAIT_FOR_UPLOAD_PROPERTY,
+] as const;
 
 let isStorageOnlyFileModeRegistryInitialized = false;
 
-function isStoreDataAsTextHidden(type: string): boolean {
-  const prop = Serializer.findProperty(type, STORE_DATA_AS_TEXT_PROPERTY);
+function isPropertyHidden(type: string, propertyName: string): boolean {
+  const prop = Serializer.findProperty(type, propertyName);
   return prop?.visible === false;
 }
 
+function areAllPropertiesHidden(type: string): boolean {
+  return HIDDEN_PROPERTIES.every((propertyName) =>
+    isPropertyHidden(type, propertyName),
+  );
+}
+
 /**
- * Hides `storeDataAsText` from the Creator property grid for file/signaturepad
- * questions. Call only when a storage provider is enabled — without one,
- * files must stay embedded as base64 text since there's nowhere to upload
- * them to.
+ * Hides `storeDataAsText` and `waitForUpload` from the Creator property grid
+ * for file/signaturepad questions. Call only when a storage provider is
+ * enabled — without one, files must stay embedded as base64 text since
+ * there's nowhere to upload them to.
  *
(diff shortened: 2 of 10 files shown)


lib/survey-features/storage-only-file-mode/infrastructure/registry.ts
```diff
@@ -0,0 +1,58 @@
...
+  STORAGE_ONLY_FILE_MODE_QUESTION_TYPES.forEach((type) => {
+    const prop = Serializer.findProperty(type, STORE_DATA_AS_TEXT_PROPERTY);
+    if (!prop) {
+      return;
+    }
+    prop.visible = true;
+  });
+
+  isStorageOnlyFileModeRegistryInitialized = false;
+}
```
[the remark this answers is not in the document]
orilux: Intentionally left as-is. `Serializer` property visibility is a global, not per-`SurveyCreatorModel` — an unregister-on-unmount would un-hide `storeDataAsText` for *every* Creator on the page, not just the one closing, and would flicker on client-side navigation between forms/templates. Storage-enabled is also a deployment-wide constant (resolved once from env vars server-side), so there's no runtime scenario where it needs to flip back. This matches the existing pattern for every other feature in `lib/survey-features/` (blind-search-tagbox, carry-forward, data-lists, question-loops) — each has a `register*Globals()` that's permanent for the session, with a `reset*ForTests()` used only in tests, never from component cleanup. What *is* correctly unregistered on unmount is the per-instance binding (`onQuestionAdded`/`onSurveyInstanceCreated` listeners), which is the part actually scoped to that creator.

lib/survey-features/storage-only-file-mode/constants.ts
```diff
@@ -0,0 +1,13 @@
+export const STORAGE_ONLY_FILE_MODE_QUESTION_TYPES = [
+  "file",
```
PeterMarinov, line 2: I would recommend that we add the `audiorecorder` question here even if it has these hard-coded in the constructor as of now and remove the hard-coded items now that we have this storage detection. Imo, this will be better. We don't have do it now, but it's good to plan for it

``` ts
constructor(name: string) {
    super(name);

    // Enforce uploading to BLOB storage
    this.waitForUpload = true;
    this.storeDataAsText = false;
    this.isRecording = false;
  }
```

lib/survey-features/storage-only-file-mode/infrastructure/survey-bindings.ts
```diff
@@ -0,0 +1,61 @@
...
+  STORAGE_ONLY_FILE_MODE_QUESTION_TYPES,
+  STORE_DATA_AS_TEXT_PROPERTY,
+  WAIT_FOR_UPLOAD_PROPERTY,
+  type StorageOnlyFileModeQuestionType,
+} from "../constants";
+
+type StorageOnlyFileModeQuestion = Question & {
+  [STORE_DATA_AS_TEXT_PROPERTY]?: boolean;
+  [WAIT_FOR_UPLOAD_PROPERTY]?: boolean;
+};
```
PeterMarinov: This type alias is redundant. I recommend we replace it with the real SurveyJS type, which is `QuestionFileModelBase`

`storeDataAsText` & `waitForUpload` are defined on `QuestionFileModelBase` - used by `audiorecorder` as well and this is what triggers the upload file event in SurveyJs as well

lib/survey-features/storage-only-file-mode/infrastructure/survey-bindings.ts
```diff
@@ -0,0 +1,61 @@
...
+
+function isStorageOnlyFileModeQuestion(
+  question: Question,
+): question is StorageOnlyFileModeQuestion {
+  return (STORAGE_ONLY_FILE_MODE_QUESTION_TYPES as readonly string[]).includes(
+    question.getType() as StorageOnlyFileModeQuestionType,
+  );
+}
```
PeterMarinov: This can become
```ts
function isFileStorageQuestion(
  question: Question,
): question is QuestionFileModelBase {
  return (STORAGE_ONLY_FILE_MODE_QUESTION_TYPES as readonly string[]).includes(
    question.getType(),
  );
}
```

PeterMarinov (dismissed):
Approved with couple of suggestions that will make the code more consistent with SurveyJS and will support future file questions

lib/survey-features/storage-only-file-mode/constants.ts
orilux: Noted

lib/survey-features/storage-only-file-mode/infrastructure/survey-bindings.ts
orilux: Addressed - removed `StorageOnlyFileModeQuestion`, now uses `SurveyJS's QuestionFileModelBase` directly.

lib/survey-features/storage-only-file-mode/infrastructure/survey-bindings.ts
orilux: Addressed - `isFileStorageQuestion` now narrows to `question is QuestionFileModelBase`, matching the suggested code.

PeterMarinov (approved):

PR #833, opened 2026-08-07
chore(h832): Upgrade Next.js to 16.3 and activate useful features

PeterMarinov:
# chore(h832): Upgrade Next.js to 16.3 and activate useful features

## Description
- Upgrade Next.js to 16.3
- Add next.js skill and upgrade React compiler to stable 1.0 version
- CI: Add file cache for builds
- Adjust app layouts for next/navigation updates
- Add new tsconfig files: tsconfig.e2e.json, tsconfig.next.json, tsconfig.vitest.json
- Update eslint config and build workflow for new Next.js requirements
- Move SEO metadata logic to lib/seo/metadata-base.ts
- Update asset storage and auth authorization tests for compatibility
- Update next.config and vitest config for Next.js 16.3 support

## Related Issues
- closes #832 

## Type of Change
Check all that apply.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Dependency update
- [ ] Refactoring (non-breaking change which improves code quality)
- [x] CI Fixes

## Checklist
- My code follows the style guidelines of this project
- I have performed a self-review of my own code
- I have commented my code, particularly in hard-to-understand areas
- I have made corresponding changes to the documentation
- My changes generate no new warnings
- I have added tests that prove my fix is effective or that my feature works
- New and existing tests pass locally with my changes
- Any dependent changes have been merged and published in downstream modules

> [!NOTE]
> - [x] I confirm that my Pull Request follows all the checklist requirements above

## Screenshots
If applicable, add screenshots to help explain your changes.

PR #834, opened 2026-08-07
fix: propagate traceparent from Hub to API via undici instrumentation

igdragoev:
# Propagate `traceparent` from Hub to API via undici instrumentation

## Description

Hub → API calls never shared a trace. Submitting a form produced a **Hub-only** trace — Aspire reported `Resources 1`, all 7 spans `endatix-hub` — while the API recorded the same request under a separate trace id of its own. There was no end-to-end view across the two services, which is what [endatix/endatix#476](https://github.com/endatix/endatix/issues/476) AC12 requires.

### Verified against a live Aspire dashboard, not reasoned about

**1. A form submission produced a Hub-only trace.** The tree contained a client span for the outbound call — `PATCH → endatix-api:8080` — but no server span from `endatix-api`:

```
POST                                              endatix-hub   ← root
└─ POST /api/public/v0/forms/[formId]/submissions  endatix-hub
   ├─ resolve page components                      endatix-hub
   ├─ executing api route (app) …                  endatix-hub
   │  └─ HTTP PATCH                                endatix-hub
   │     └─ PATCH → endatix-api:8080               endatix-hub   ← outbound, but no API span follows
   └─ start response                               endatix-hub
```

**2. The API was ruled out as the cause.** Calling it directly with an explicit header:

```
traceparent: 00-abcdef00000000000000000000000001-1111111111111111-01
```

produced a trace under *exactly that id* containing both requests. So the API honours incoming context correctly — the fault was entirely Hub-side.

### Cause

Node 18+ global `fetch` is **undici**, which bypasses `node:http` entirely, so `HttpInstrumentation` never sees these calls. `FetchInstrumentation` is the **browser** instrumentation and does nothing inside a `NodeSDK`.

Next.js *did* trace the calls through its own fetch instrumentation — which is why client spans existed and the problem looked subtler than it was — but that creates a span **without injecting a `traceparent` header**, so the API started a fresh trace every time.

`UndiciInstrumentation` patches the transport the Hub actually uses (`lib/endatix-api/endatix-api.ts:285,313` call global `fetch`) and injects the header.

## Related Issues

Refs endatix/endatix#476

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Dependency update
- [ ] Refactoring
- [ ] Other

## Checklist

> [!NOTE]

## Verification

- `features/telemetry` — **57 tests pass** (7 files)
- The new test is a **real guard**, not a vacuous assertion: with the `new UndiciInstrumentation()` line removed it fails with `expected [ …(2) ] to include '@opentelemetry/instrumentation-undici'`. The `(2)` confirms the array is genuinely populated (http + fetch) rather than the check passing on an empty list.
- `tsc --noEmit` reports no new errors. The existing ones are all in `__tests__/**` and `components/user/user-avatar.tsx`, none in `features/telemetry` — and #833 adds `tsconfig.next.json` precisely to keep tests out of the build typecheck.

> [!IMPORTANT]
> **What I could not verify:** that a form submission now produces a single trace with `Resources 2`. That needs a running Hub, API and dashboard together. The cause is evidenced above and the mechanism is well understood, but the end-to-end result is unconfirmed — please check the dashboard after this lands and look for the API span nested under the Hub span.

## Notes for the reviewer

- **The version number looks odd on purpose.** `@opentelemetry/instrumentation-undici` is on its own line at `0.31.0`, not the `0.221.0` shared by the sibling packages. It is not out of date — its only peer dependency is `@opentelemetry/api ^1.7.0`, satisfied by the `^1.9.1` already present.
- **The test reads a private `_instrumentations` field.** Fragile across a major OTel bump, but the SDK exposes no public accessor, and a guard that might need revisiting beats no guard for a regression this quiet — it produced *plausible-looking* traces the whole time it was broken.
- **`FetchInstrumentation` is deliberately left in place.** It is inert on the server and removing it is defensible cleanup, but that is a separate judgement from fixing AC12 and would muddy this diff.
- **Relationship to #833 (Next.js 16.3):** no overlap. That PR touches no telemetry or instrumentation files; the only shared file is `pnpm-lock.yaml`, which will conflict trivially — take theirs and re-run `pnpm install`. I'd suggest merging #833 first and rebasing this on top, since it is far the larger change. The diagnosis was made on 16.2.12, so re-checking the trace after the upgrade is cheap insurance.

### DIFF
--- a/features/telemetry/__tests__/otel-telemetry-strategy.test.ts
+++ b/features/telemetry/__tests__/otel-telemetry-strategy.test.ts
@@ -32,6 +32,28 @@ describe("OtelTelemetryStrategy", () => {
     expect(strategy.name).toBe("OTel");
   });
 
+  it("should register undici instrumentation so traceparent reaches the API", () => {
+    // Arrange — Node 18+ global fetch is undici and bypasses node:http, so without this
+    // instrumentation no traceparent is injected and Hub -> API calls land in separate traces.
+    // Verified against a live Aspire dashboard: form submissions produced a Hub-only trace
+    // ("Resources 1") until this was added.
+    process.env.OTEL_EXPORTER_OTLP_ENDPOINT = "http://localhost:4317";
+    const strategy = new OtelTelemetryStrategy();
+
+    // Act
+    const sdk = strategy.initialize(resource);
+
+    // Assert
+    const registered = (
+      sdk as unknown as { _instrumentations?: unknown[] }
+    )._instrumentations;
+    const names = (registered ?? []).map(
+      (i) => (i as { instrumentationName?: string }).instrumentationName,
+    );
+
+    expect(names).toContain("@opentelemetry/instrumentation-undici");
+  });
+
   it("should throw an error if the endpoint is not configured", () => {
     // Arrange
     process.env.OTEL_EXPORTER_OTLP_ENDPOINT = undefined;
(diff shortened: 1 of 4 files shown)


PeterMarinov (approved):
Great findings. Can't wait to test it

PR #850, opened 2026-08-12
fix(h829): Carry forward drops values selected from a lazy-loaded data list past the first page

### LINKED ISSUE #829: Carry forward drops values selected from a lazy-loaded data list past the first page
### Summary

Carry forward (`edxCarryForwardEnabled`) silently fails to carry values from a source question whose choices come from a data list (`edxDataListId` + `choicesLazyLoadEnabled`) when the selected value lies beyond the first lazy-loaded page (default page size: 25). Depending on timing, the value either never appears in the target's choices, or appears for a moment and then disappears.

### Steps to Reproduce

1. Build a survey with a lazy-loaded data-list source and a carry-forward target:

   ```json
   {
     "pages": [
       {
         "name": "page1",
         "elements": [
           {
             "type": "tagbox",
             "name": "question1",
             "choicesLazyLoadEnabled": true,
             "edxDataListId": "1534969877259878400"
           },
           {
             "type": "checkbox",
             "name": "question2",
             "choices": ["Item 1", "Item 2", "Item 3"],
             "edxCarryForwardEnabled": true,
             "edxCarryForwardSources": ["question1"],
             "edxCarryForwardMode": "selected"
           }
         ]
       }
     ]
   }
   ```

2. Point the data list at 47 alphabetically-sorted country names (Algeria … Uzbekistan).
3. In `question1`, select a country ranked **# 25 or earlier** (e.g. "Jordan"). It correctly appears as a choice in `question2`.
4. In `question1`, select a country ranked **# 26 or later** (e.g. "Mexico"). It does **not** reliably appear in `question2` — either it never shows up, or it flashes in for a moment and then disappears.

### Expected Behavior

Any value selected in the source question is carried forward to the target's choices, regardless of whether the source question happens to have that item's page loaded locally at sync time.

### Actual Behavior

- Selections from unloaded pages are missing from the target's choices entirely, **or**
- They appear briefly (while a search request transiently includes that item in the source's local choices) and then vanish once the source's dropdown reloads its default first page.

### Root Cause

The boundary (item 25 vs 26) is not a coincidence — it's the default lazy-load page size.

1. **The data list is only ever partially loaded client-side.** `choicesLazyLoadPageSize` defaults to 25 (`hub/lib/survey-features/data-lists/constants.ts:3`, matching SurveyJS's own tagbox default). The 47-item list splits into "Algeria…Jordan" (page 1, items 1–25) and "Mexico…Uzbekistan" (page 2+, items 26–47) — exactly the reported cutoff.

2. **SurveyJS overwrites, not accumulates, that page.** Every time the tagbox's dropdown opens or closes, `DropdownListModel.resetItemsSettings()` clears its item cache and refetches from `skip: 0` (survey-core: `dropdownListModel.ts`, `resetItemsSettings`/`setItems`). Whatever the question's `choices` array held from a prior search is discarded and replaced with just the newly-fetched page.

3. **SurveyJS's usual safety net is explicitly disabled for lazy load.** Normally a selected value with no matching item in `choices` gets a synthetic fallback via `customChoices`, but tagbox skips that specifically when `choicesLazyLoadEnabled` is true (survey-core: `question_tagbox.ts`, `updateCustomChoices`: `&& !this.choicesLazyLoadEnabled`). So there's no built-in mechanism to remember a selection whose page has since been evicted.

4. **Carry forward reads only the currently-loaded page.** `getChoicesFromSourceQuestion` (`hub/lib/utils/survey/get-choices-from-source-question.ts`) computed its output purely from `source.visibleChoices` — which, per points 2–3, only reflects whatever page is loaded *right now*. A selected value outside that page was invisible to it.

5. **The sync actively deletes stale carried values.** `syncSingleCarryForwardTarget` (`hub/lib/survey-features/carry-forward/use-cases/sync-carry-forward-target.ts`) calls `target.clearIncorrectValues()` whenever the computed choice list changes. Once the source's loaded page no longer includes the selection, the next sync recomputes a choice list without it and prunes it from the target — this is the "appears, then disappears" flicker.

This isn't Endatix-specific, either — SurveyJS's own native `choicesFromQuestion` carry-forward (`question_baseselect.ts`, `getChoicesFromSelectQuestion`) has the identical gap. Endatix's implementation was modeled on it and inherited the limitation; lazy-loaded/data-list sources are an Endatix-specific extension on top of that primitive, so upstream never had to handle this case.

### Proposed Fix

Modify `getChoicesFromSourceQuestion` (`hub/lib/utils/survey/get-choices-from-source-question.ts`) so it does not rely on `visibleChoices` alone when the source is lazy-load-enabled and the mode is "Selected Only":

- After building the result from `visibleChoices`, also read the source's raw `value`.
- For any selected value with no matching entry in the currently-loaded choices, synthesize a fallback `ItemValue` via `source.createItemValue(value)`, using the raw value as its text.
- Exclude none/other/refuse/don't-know sentinel values from this fallback — handle those through the existing separate logic, not this path.
- Append the synthesized fallback items to the result, so a selection is never dropped purely because its page isn't loaded at sync time.

Scope the change tightly:

- Only trigger the fallback for lazy-load sources (check `"choicesLazyLoadEnabled" in source`). Leave static-choice carry-forward behavior untouched.
- Only trigger it in "Selected Only" mode. Leave "All" and "Unselected Only" alone — both are inherently unbounded for a remote list, so there's no complete set to fall back to.
- Keep the fix fully synchronous. Do not introduce API calls or async logic here — the existing reentrancy-guarded sync path (`updatingCarryForwardSurveys`) assumes synchronous execution.

**Flag, don't fix in this pass:** the fallback item's label will be the raw stored value, not a resolved display label. This is invisible when a data list's value and label are identical strings, but for data lists with distinct value/label columns, a carried-over item will show the raw value until something re-resolves it. If that turns out to matter, resolve it by wiring in the existing API-backed resolver (`resolveDataListDisplayValues`, already used by the property-grid and PDF-export paths) — that requires making `syncSingleCarryForwardTarget` async, so treat it as a separate, larger follow-up rather than bundling it into this fix.

PeterMarinov:
# fix(h829): Carry forward drops values selected from a lazy-loaded data list past the first page

## Description
- Improved carry-forward choices when source data is lazy loaded.
- Preserved selected values and localized display labels, including selections outside the currently loaded page.
- Added more reliable multilingual label resolution for selected choices.
- Prevented resolved labels from being lost during refreshes or partial data updates.
- Improved handling of incomplete, delayed, or stale label responses.
- Preserved unchanged choice data to avoid unnecessary updates.
- Added fallback labels when localized display data is unavailable.
- Ensured dependent selections update after newly selected values are resolved.
- Refactor duplicated code in data list create/replace
- Refactor result pattern to allow for easy access in same exported "@lib/result" namespace

## Related Issues
- closes #829 

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [x] Refactoring (non-breaking change which improves code quality)
- [ ] Other (please describe)

## Checklist

> [!NOTE]

## Screenshots
If applicable, add screenshots to help explain your changes.

## Summary by CodeRabbit

* **New Features**
  * Improved carry-forward choices when source data loads incrementally.
  * Preserved selected values and localized display labels, including selections outside the currently loaded page.
  * Added more reliable multilingual label resolution for selected choices.

* **Bug Fixes**
  * Prevented resolved labels from being lost during refreshes or partial data updates.
  * Improved handling of incomplete, delayed, or stale label responses.
  * Preserved unchanged choice data to avoid unnecessary updates.
  * Added fallback labels when localized display data is unavailable.
  * Ensured dependent selections update after newly selected values are resolved.


PeterMarinov (commented):
Checked the code

orilux (dismissed):

orilux (approved):

PR #868, opened 2026-08-19
feat(h865): add URL-driven list grid with search and locale filter

PeterMarinov:
## Summary
  - Data lists management grid is URL-driven  (`search`, `hasLocale`, paging) and uses the  shared table chrome.
  - Hub list client now sends API `search` (not   `query`) on `GET /data-lists`, matching  forms/users. Item/public search still uses   `query`.
  - Locale filter on list URLs is normalized the  same way for grid links, create-dialog close, and  delete.
  - Fixed: "Back to Data Lists" on the detail page now restores the list's filters/paging via `from`/`parseDataListsReturnHref` instead of  always returning to an unfiltered `/data-lists`.
  - Fixed: the search and locale filters now share a single debounced URL update so editing both in quick succession can't drop one of the changes.
  - Docs: corrected project-structure.md — the list еndpoint's API param is `search` (not `query`), and `hasLocale` takes a single locale code  (comma-separated OR isn't implemented).
  - Cleanup: de-duplicated the `firstString` search-param helper and the list-query paging normalization.

image

  ## Test plan
  - [x] List page: search by name/description;  confirm network query string uses `search=`
  - [x] Filter by locale (`es`); confirm  `hasLocale=es` and invalid input is dropped
  - [x] Open a list with filters/paging set, open a  data list, click "Back to Data Lists" -  filters/paging restore
  - [x] Type in the search box and the locale box  in quick succession; confirm both end up in the URL (neither is dropped)
  - [ ] Open create via `?action=create`, close dialog; `action` is removed, filters stay
  - [x] Item grid search still uses `query` (not  `search`)
  - [x] `view-lists/__tests__/list-data-lists-query .test.ts`, `lib/endatix-api/data-lists/__tests__/list-data-lists-endpoint.test.ts`,  `components/table/__tests__/*`


PeterMarinov (commented):
Reviewed

orilux (approved):

PR #903, opened 2026-08-26
feat(h876): add assume-tenant session swap and support banner

PeterMarinov:
## Summary
Platform admins can **assume a tenant for support** from `/admin/tenants` (confirm first), work in that tenant context, and **exit** via a sticky banner.

- Row action **Assume tenant** → confirm dialog → session cookie swap → redirect to `/forms`.
- Sticky **Support access** banner in the main shell while JWT has `act` + `tid`; **Exit tenant** restores the prior session and returns to `/admin/tenants`.
- Tenants list uses the shared Hub data table (search, sort, date filters, paging). Create/edit no longer collect an auth-provider allow-list (API still accepts the field; Hub omits it).
-  Tenant creation now uses a guided three-step wizard with a review screen.
-  Added reusable status indicators, panel sections, progress steps, and summary rows.

## APIs
- `POST /auth/assume-tenant` `{ tenantId }` → `{ accessToken, refreshToken }`
- `POST /auth/exit-assume` → `{ accessToken, refreshToken }`
- Existing `GET /admin/tenants/{id}` used only to label the banner.

Actions go through `requireTenantManagement` + `toResult`. Token swap is `replaceSessionTokens` (`server-only`, not a client-callable action).

## Events
None from Hub.

## Key decisions
- Assume is **support access**, not membership. Banner copy states that.
- Confirm before assume; dismiss on the banner is visual-only (session stays assumed until Exit).
- After assume, land on the tenant **forms list**, not the tenants admin page.
- Gate with the same `multi-tenancy` flag as create/edit.

## Behavior / breaking
- Additive Hub UI + auth client methods. Requires API assume/exit endpoints.
- Hub create/edit stop sending `allowedAuthProviderKeys` (empty/default on the API).
- Not a list-envelope wire break; tenants list now uses `normalizePagedResponse` client-side.

### Screenshots
image
image
image
image

## Summary by CodeRabbit

* **New Features**
  * Support staff can enter and exit tenant-support sessions, with a visible banner identifying the active tenant.
  * Added tenant list search, sorting, date filters, pagination, loading states, and tenant assumption actions.
  * Tenant creation now uses a guided three-step wizard with a review screen.
  * Added reusable status indicators, panel sections, progress steps, and summary rows.

* **Improvements**
  * Improved validation, alerts, accessibility, and responsive table presentation.
  * Removed authentication-provider configuration from tenant creation and editing.

* **Documentation**
  * Updated guidance for tenant support sessions, list pages, overlays, and status displays.


lib/list-page/use-list-url-state.ts
```diff
@@ -20,3 +18,5 @@ export function useListUrlState(
 
   return { search, setSearch, urlSearch, updateUrl, searchParams, isPending };
 }
+
+export type ListUrlState = ReturnType<typeof useListUrlState>;
```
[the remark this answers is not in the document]
PeterMarinov: This is done. Check

features/platform-admin/update-tenant/ui/edit-tenant-sheet.tsx
```diff
@@ -131,26 +126,29 @@ export function EditTenantSheet({
       <ResponsivePanelHeader>
         <ResponsivePanelTitle>Edit tenant</ResponsivePanelTitle>
         <ResponsivePanelDescription>
-          Update the display name and self-registration policy. The public id
-          stays locked because sign-in URLs already use it.
+          Update the name, description, or self-registration policy.
         </ResponsivePanelDescription>
       </ResponsivePanelHeader>
 
       {isLoading || !shortUrl ? (
```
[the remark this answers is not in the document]
PeterMarinov: This is fixed. Check

features/platform-admin/assume-tenant/ui/support-access-banner-view.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <AlertTitle className="col-start-auto min-h-0 shrink-0">
+          {title}
+        </AlertTitle>
+        <AlertDescription className="col-start-auto line-clamp-2 md:line-clamp-1">
+          You are viewing this tenant as a platform administrator. You are not a
+          member of the tenant.
+        </AlertDescription>
+      </div>
+      <AlertAction className="col-start-2 mt-1 flex items-center gap-1 md:col-start-3 md:row-start-1 md:mt-0">
+        <form action={exitAction}>
```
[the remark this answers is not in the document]
PeterMarinov: This is fixed. Check

features/platform-admin/assume-tenant/replace-session-tokens.ts
```diff
@@ -0,0 +1,29 @@
...
+    await unstable_update({
+      accessToken,
+      refreshToken,
+      expiresAt: payload.exp || Date.now() / 1000,
+    });
+
+    if (payload.sub) {
+      invalidateUserAuthorizationCache({ userId: payload.sub });
+    }
+
```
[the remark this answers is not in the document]
PeterMarinov: This is done. Validate

components/table/data-table-column-meta.ts
```diff
@@ -0,0 +1,9 @@
+import "@tanstack/react-table";
+
+declare module "@tanstack/react-table" {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  interface ColumnMeta<TData, TValue> {
+    cellClassName?: string;
+    headerClassName?: string;
+  }
```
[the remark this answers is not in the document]
PeterMarinov: This is done. Check and confirm

app/(main)/layout.tsx
```diff
@@ -59,6 +62,7 @@ export default async function RootLayout({
         >
           {nav}
           <main data-slot="sidebar-inset">
+            <AssumedTenantBanner accessToken={session?.accessToken} />
```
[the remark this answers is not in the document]
PeterMarinov: This is done and moved to separate wrapper component

features/platform-admin/assume-tenant/assume-tenant.action.ts
```diff
@@ -0,0 +1,72 @@
...
+  );
+  if (Result.isError(result)) {
+    return result;
+  }
+
+  return swapSessionAndRedirect(result.value, "/forms");
+}
+
+export async function exitAssumeAction(): Promise<ResultType<boolean>> {
+  const api = await requireTenantManagement();
```
igdragoev: **Exit is gated on the same guard as enter, so a gate change mid-session strands the operator.** `requireTenantManagement()` is PlatformAdmin authorization *plus* the multi-tenancy flag, and both can change while a session is assumed:

- Flag turned off → exit returns `Result.error("Tenant management is not enabled for this environment.")` and the operator stays in tenant 42.
- PlatformAdmin approval revoked → `requirePlatformAdmin()` **throws** rather than returning a `Result` (see the doc comment on `requireTenantManagement`), so `SupportAccessBannerView`'s `Result.isError` toast never runs and the action rejects into an error boundary.

Either way the only route back to the home tenant is a full sign-out. Exit is session recovery, not tenant management — it should be gated on `readAssumeSession(session.accessToken) !== null` alone, and call `/auth/exit-assume` with the current session token.

igdragoev (commented):
Six comments, all new — the CodeRabbit threads already resolved on this PR aren't re-raised. The first two are the ones I'd want resolved before merge (session lockout and a cross-tenant cache path); the rest are correctness issues in the new parsing/session code.

features/platform-admin/assume-tenant/assume-tenant.action.ts
```diff
@@ -0,0 +1,72 @@
...
+): Promise<ResultType<boolean>> {
+  const swapped = await replaceSessionTokens(
+    tokens.accessToken,
+    tokens.refreshToken,
+  );
+  if (Result.isError(swapped)) {
+    return swapped;
+  }
+
+  redirect(href);
```
igdragoev: **The swap replaces the session cookie but never invalidates Next's route/router caches.** Repro: admin visits `/forms` in their home tenant → `/admin/tenants` → assumes tenant 42 → lands back on `/forms`. Any segment already in the client Router Cache (back/forward, a prefetched `<Link>`, a statically-cached segment) can render the **home tenant's** forms inside the assumed session.

That's a cross-tenant data exposure in the one feature whose banner promises tenant isolation. `swapSessionAndRedirect` should `revalidatePath("/", "layout")` before `redirect(href)`, which covers both the assume and exit paths from one place.

features/platform-admin/list-tenants/utils.ts
```diff
@@ -0,0 +1,106 @@
...
+    request.page ?? 1,
+    request.pageSize ?? DEFAULT_TENANTS_PAGE_SIZE,
+    request.search,
+    request.sortBy,
+    request.sortDir,
+    request.createdFrom,
+    request.createdTo,
+    request.modifiedFrom,
+    request.modifiedTo,
+  ].join("|");
```
igdragoev: **The Suspense key can collide across different queries, so the boundary doesn't remount.** Raw user input is joined with the same `"|"` used as the separator: search `acme|name` with no sort gives `1|10|acme|name|||||`, byte-identical to search `acme` sorted by `name`.

Switching between those two states keeps the boundary mounted, so `use(tenantsPromise)` hands back the previous render's already-settled result and the table shows the wrong tenants. `encodeURIComponent` each segment, or just `JSON.stringify(request)`.

features/platform-admin/assume-tenant/replace-session-tokens.ts
```diff
@@ -0,0 +1,42 @@
...
+  accessToken: string,
+  refreshToken: string,
+) {
+  let payload: EndatixJwtPayload;
+  try {
+    payload = decodeJwt<EndatixJwtPayload>(accessToken);
+    await unstable_update({
+      accessToken,
+      refreshToken,
+      expiresAt: payload.exp ?? Date.now() / 1000,
```
igdragoev: **`payload.exp ?? Date.now() / 1000` makes an `exp`-less token expire immediately**, and writes a non-integer epoch where every other `expiresAt` in the session is integer seconds.

If `/auth/assume-tenant` ever returns a token without `exp`, the session is written with `expiresAt` = now; the jwt callback in `config-factory.ts` stamps `error: "SessionExpiredError"` whenever `expires_at - 10s < now`, so the first request after the redirect signs the operator out instead of showing the assumed tenant.

A missing `exp` is a malformed token — better as a `Result.error` than a fallback. If a fallback stays, it needs to be `Math.floor(Date.now() / 1000) + <ttl>`.

features/auth/authorization/domain/permissions.ts
```diff
@@ -9,6 +9,7 @@ export const Permissions = {
   },
   Platform: {
     ManageTenants: "platform.tenants.manage",
+    AssumeTenants: "platform.tenants.assume",
```
igdragoev: **`AssumeTenants` is added but never read.** Assume and exit are gated on `requireTenantManagement()` — PlatformAdmin role + multi-tenancy flag — not on this permission, so the Hub gate is looser than the API's: the OSS API declares `platform.tenants.assume` as distinct from `platform.tenants.manage`, and an admin whose role grants manage-but-not-assume still sees and can click **Assume tenant**, failing only at the API.

Either check it in `assumeTenantAction`, or drop the constant until the slice that uses it.

features/platform-admin/utils.ts
```diff
@@ -25,18 +21,6 @@ export function parsePlatformAdminListParams(
   };
```
igdragoev: **The sibling parser in this same file didn't get the hardening.** The tenants parser moved to `list-tenants/utils.ts` with `firstSearchParam` / `SearchParamValue` (and project-structure.md now documents that as the rule), but `parsePlatformAdminListParams` above still does `searchParams?.search?.trim()` and `searchParams?.tenantId?.trim()` on fields typed as `string` (lines 18 and 20).

`/admin/platform-admins?search=a&search=b` — Next delivers `search` as `string[]`, `.trim is not a function` throws, and the route hits `error.tsx`. `PlatformAdminSearchParams` in `features/platform-admin/types.ts` (also edited in this PR) still declares `search?: string`, so TypeScript doesn't catch it. Same fix as tenants: `SearchParamValue` members plus `firstSearchParam`.

features/platform-admin/assume-tenant/assume-tenant.action.ts
PeterMarinov: Agreed for Hub. Exit is session recovery, not tenant CRUD. It now uses the current session token and readAssumeSession — no requireTenantManagement, so a Hub flag flip or requirePlatformAdmin throw cannot block Exit. Assume still uses that gate. Note: if the API multi-tenancy flag is off, OSS still 404s on exit-assume; that’s not fixable in this Hub PR

features/platform-admin/assume-tenant/assume-tenant.action.ts
PeterMarinov: Good catch. After a successful cookie swap we now revalidatePath("/", "layout") before redirect so assume and exit both drop stale layout/route cache.

features/platform-admin/list-tenants/utils.ts
PeterMarinov: Agreed. tenantsListSuspenseKey is now JSON.stringify of the list request, so a | in search cannot collide with sort.

features/platform-admin/assume-tenant/replace-session-tokens.ts
PeterMarinov: Agreed. A token without numeric exp is treated as malformed — we return an error and do not write the session cookie.

features/auth/authorization/domain/permissions.ts
PeterMarinov: Won’t check or drop it. Endatix API has  `Actions.Platform.AssumeTenants` is reserved and not catalog-seeded; as v1 of the feature is for platform admins only. Thus, the API gate is PlatformAdmin + act, not this permission. Hub mirrors that catalog (same as ImpersonateUsers). 

The permission is reserved and documented in agent docs and code so it's used once this permission ca be granted to other users

features/platform-admin/utils.ts
PeterMarinov: Agreed. parsePlatformAdminListParams now uses firstSearchParam / SearchParamValue the same way as tenants, so repeated query keys don’t crash the route.

orilux (approved):

igdragoev (approved):

Issue #920, opened 2026-08-31
Move inlined config values to runtime env


PeterMarinov:
### Goal

Operators using Docker Compose or Kubernetes must be able to change SurveyJS license, reCAPTCHA, PostHog, debug mode, and submitter display labels **without rebuilding the Hub image**. Next.js inlines `NEXT_PUBLIC_*` at `next build`; the same applies to keys merged via `nextConfig.env`. Hub config must be read at **request time** and projected to the browser through an explicit allowlist.

### Implementation Notes

- **PR 917** ([feat/hub-runtime-client-config](https://github.com/endatix/endatix-hub/pull/917)): renames `NEXT_PUBLIC_*` → `ENDATIX_*`, Helm wiring, legacy boot normalisation via `applyLegacyPublicEnv()`, SurveyJS license scoped off the shared public projection.
- **Follow-up PR** (`feat/runtime-config-request-time-api`): config-safe vs request-time barrels (`@/features/config` vs `@/features/config/server`), `EndatixResolvedSettings.client` slice, `await connection()` inside `getClientEndatixConfig()`, `survey-license.server.ts`, docs and leak guards.

### Out of scope (later PRs)

- Helm `secretKeyRef` for license / PostHog key
- PostHog `/ingest` rewrites via `withEndatix`
- SaaS CI moving keys from Hub build `env:` to runtime
- Deleting `legacy-public-env.server.ts` after deprecation window

PR #928, opened 2026-09-03
docs: add config guidance to AGENTS.md, point CLAUDE.md at it

igdragoev:
## Description

`AGENTS.md` had nothing on configuration, which is the area that most reliably produces wrong diagnoses: a container env var that "has no effect", or a fix reached for with `--build-arg` that quietly forces one image per environment.

#917 settled where the line falls. This records it so the next person — or agent — does not have to re-derive it from `features/config/`:

- Public values resolve at **request time** via `getClientEndatixConfig()`; the `ENDATIX_*` prefix replaces `NEXT_PUBLIC_*` for anything public.
- The **prefix**, not the `"use client"` / `"use server"` boundary, is what decides whether a value gets baked into the bundle.
- Adding a public value touches three places and lands in the HTML of every page mounting `AppProvider`, anonymous form layouts included — so a secret gets scoped to its routes instead, the way `SurveyLicenseProvider` does.
- `basePath` is the one remaining build-time exception, which is why subfolder hosting needs per-origin hosting or a self-build.

`CLAUDE.md` is added as a five-line pointer so `AGENTS.md` stays the single source. Every other repo in the workspace already keeps its guidance in `AGENTS.md`; the Hub had an untracked `CLAUDE.md` living on one machine, which is what prompted this.

## Related Issues

Follows #917. Refs #920.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Dependency update
- [ ] Refactoring
- [ ] Other

## Checklist

> [!NOTE]

## Notes for the reviewer

- Docs only — no code touched.
- The section is deliberately short (29 lines) and duplicates no code. It states the rules and names the symbols; `features/config/` and its guard tests remain the detail.
- Worth confirming one claim reads right: that `AssertNoSecretsInClientConfig` "only backstops a few known secret names" — it lists four, so it is a safety net rather than a general check, and I wanted the doc to say so rather than imply broader coverage.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


PeterMarinov (approved):
Approved

PR #861, opened 2026-08-13
feat(h842): Implemented height mode for embedded forms

orilux:
## Description

Implements `data-height-mode="fill"` for the embed SDK. Customers embedding a form inside a fixed-height container (sidebar, modal) previously got a visible gap below the iframe, because the SDK only ever sized the iframe to the form's content height. This adds an opt-in `fill` mode that makes the iframe fill the container when content is shorter, and still grow beyond it when content is taller — with `auto` (unchanged) staying the default.

**Approach:** the issue's own suggested implementation (`ResizeObserver` + `Math.max(contentHeight, measuredContainerHeight)` in JS) has a real bug for its own documented edge case: when the parent container has no explicit height, `clientHeight` just mirrors the iframe's own previous rendered height, so the JS-computed max never lets the iframe shrink back down once it's grown — a permanent ratchet. Instead, this uses pure CSS: the wrapper div gets `height: 100%` and the iframe gets `min-height: 100%`, while the existing content-driven `iframe.style.height` logic is untouched. The browser's native `max(height, min-height)` resolution handles "fill or grow, whichever is bigger" on every layout pass with no JS measurement, and CSS percentage-height only resolves against an ancestor with a *definite* height — so a container with no explicit height (or only `min-height`) falls back to exactly today's `auto` behavior automatically, with zero custom detection code.

**Also fixed:** verifying visually surfaced a second issue beyond the outer iframe box — the survey's own document (`html`/`body` inside the iframe) didn't stretch or paint to match the now-larger iframe, leaving a transparent gap that showed the *host page's* background through it instead of the survey's. Fixed by propagating `heightMode` into the iframe via a query param (gated on `embedId` also being present, so a manually-opened embed URL doesn't trigger it) and, in fill mode, painting `html`/`body` with the survey's own theme background — with a fallback chain for missing theme variables, a default applied immediately server-side to cover the loading window, and a fix for a render race where the effect could get stuck on the pre-theme `DefaultLight` color.

### Changes
- `src/embed/embed.ts` — `data-height-mode` attribute / `heightMode` option, CSS-based fill logic, `parseHeightMode` validator. Also fixes a pre-existing bug (unrelated to height mode, but made more visible by it) where a container could be left orphaned in the DOM if base URL resolution failed.
- `features/embed-form/height-mode.ts` — shared `isFillHeightMode` check + default background color, used by both the server-rendered embed page and the client messaging context so there's one source of truth.
[...]

PR #936, opened 2026-09-06
fix(h930): Background color not filling height in embedded survey under `fill` mode

### LINKED ISSUE #930: Unexpected background color on SurveyJS 3.0 on some legacy themes
## Description

SurveyJS 3.0 has introduced a new default background color on surveys with legacy themes. This color creates an unexpected color change, as well as a separate color section when the form is embedded with `data-heigh-mode=fill`.

## Screenshot

Image

orilux:
## Description

Fixes a regression from the SurveyJS 3.0 upgrade: embed fill-mode's background-fill (added in #861 to fix h842) stopped working, leaving a gray gap instead of the survey's own theme color.

**Root cause:** SurveyJS v3 no longer exposes the page/surface background as a CSS custom property readable off `themeVariables` and applicable via inline `style` — it injects its own `:where(.sd-theme-root)` stylesheet instead, which wins the cascade regardless of which `--sjs*`/`--sjs2*` variable name is guessed. The color still exists, just on a pseudo-element (`.sd-root-modern::before`) that isn't readable via `getPropertyValue` on any real element.

**Fix:** read the browser's own resolved color directly off `.sd-root-modern::before` via `getComputedStyle`, instead of guessing which variable name this SurveyJS version uses — version-agnostic, since it asks what SurveyJS actually rendered rather than replicating its internal variable resolution. Also fixes a related timing bug: the paint effect could previously fire before `isModelReady`, i.e. before `<Survey>` had rendered `.sd-root-modern` into the DOM at all.

**Hardening from review:** added a listener on SurveyJS's own `onAfterRenderSurvey` event as a safety net, since this component's effect timing relative to `survey-react-ui`'s internal rendering isn't a documented contract either side promises. Also wired the CSS loading-phase fallback color to the actual `DEFAULT_FILL_BACKGROUND_COLOR` constant (via a custom property) instead of a second hardcoded literal that could drift out of sync.

### Testing
- Unit: real `.sd-root-modern` + stubbed `::before` in the test mock (previously the mock had neither, so every test silently only exercised the fallback path); new cases for the primary DOM-read path, the storage-not-ready→ready transition, and the `onAfterRenderSurvey` safety net.
- E2E: background-color assertion is now `expect.poll`-based (was a single-shot check) and asserts equality with the survey's own rendered color, not just "not transparent" — the previous version of this exact assertion is what let the regression through unnoticed.
- Verified live against the real embed flow, including the grow-then-shrink case from the original ratchet fix, which still holds.

## Related Issues
#930 

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [ ] Other (please describe)

## Checklist

> [!NOTE]

## Summary by CodeRabbit

- **Bug Fixes**
  - Embedded forms with shorter content now correctly fill fixed-height parent containers.
  - The surrounding background now matches the rendered survey theme more reliably.
  - Background painting waits for the survey to be ready and refreshes after rendering, reducing flashes of incorrect or transparent colors.
  - A neutral fallback background is used when the theme color cannot be detected.


e2e/tests/embed/embed-form.spec.ts
```diff
@@ -212,11 +212,40 @@ test.describe("Embed Form Height Modes (Real Environment)", () => {
...
+
+    // Poll, not a single read: the background is painted in a useEffect
+    // that can run after .sd-root-modern first becomes visible, so a
+    // one-shot check here would be flaky (still transparent) rather than
+    // a reliable signal either way.
+    await expect
+      .poll(async () => {
+        const colors = await readColors();
+        return colors.body;
+      })
```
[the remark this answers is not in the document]
orilux: Addressed

features/public-form/ui/survey-component.tsx
```diff
@@ -121,51 +148,62 @@ export default function SurveyComponent({
...
+      const pseudoBackground = card
+        ? getComputedStyle(card, "::before").backgroundColor
+        : "";
+      const themeVariables = surveyModel.themeVariables ?? {};
+      const backgroundColor = isPaintedColor(pseudoBackground)
+        ? pseudoBackground
+        : themeVariables["--sjs-general-backcolor-dim"] ||
+          themeVariables["--sjs-general-backcolor"] ||
+          DEFAULT_FILL_BACKGROUND_COLOR;
+
```
PeterMarinov: _🎯 Functional Correctness_ | _🟡 Medium_

**Drop the `themeVariables` / `--sjs-*` fallback now that 3.x is the floor.**

This branch still reads `--sjs-general-backcolor-dim` / `--sjs-general-backcolor` when `::before` is not painted yet. That is the v2-style guess this PR is replacing, and v3 does not expose those as a reliable canvas color on our wrapper.

If `::before` is empty, consider leaving html/body alone (the CSS `--embed-fill-fallback-bg` already covers the brief window) and painting only once `onAfterRenderSurvey` has a real computed color.

Keeping `DEFAULT_FILL_BACKGROUND_COLOR` as a last resort is fine; the `--sjs-*` lookups are what look leftover from pre-3.x.

features/public-form/ui/survey-component.tsx
```diff
@@ -121,51 +148,62 @@ export default function SurveyComponent({
...
-    formId,
-    getSubmissionId,
-    onSubmissionIdChange: handleSubmissionIdChange,
-  });
-
-  const isModelReady = surveyModel && isStorageReady;
+  }, [isFillMode, isModelReady, surveyModel, appliedTheme]);
 
   const { sendEmbedMessage, registerEmbedHandlers } = useSurveyEmbedBehavior({
     isEmbed: isEmbed ?? false,
```
PeterMarinov: _🐛 Possible Bug_ | _🟡 Medium_ | _⚠️ Needs verification_

**Cleanup may flash the host page when `appliedTheme` changes.**

`previousHtmlBackground` / `previousBodyBackground` are captured at effect start (often empty). `appliedTheme` is in the dep list, so a later theme apply runs cleanup first and restores empty inline styles, then paints again.

I have not reproduced this in the embed iframe — please check fill + a stored (non-default) theme: any transparent/host-bleed frame between theme JSON landing and the second paint?

If yes, keep the original canvas colors in a ref and restore only on unmount / leaving fill mode, not on every `appliedTheme` tick.

e2e/tests/embed/embed-form.spec.ts
```diff
@@ -212,11 +212,43 @@ test.describe("Embed Form Height Modes (Real Environment)", () => {
...
+    // theme, or this passes even with a hardcoded/wrong fallback color
+    // (exactly how h930's SurveyJS-3.0 regression slipped through: this
+    // assertion previously only checked for "not transparent"). Note this
+    // only proves the fallback is unused for *this* seeded form, which has
+    // no stored theme (DefaultLight throughout, no override to race
+    // against) — see survey-component.test.tsx for coverage of a theme
+    // arriving after the fallback has already painted.
+    expect(colors.cardSurface).toBeTruthy();
+    expect(colors.body).toBe(colors.cardSurface);
   });
```
PeterMarinov: _🧪 Test Coverage_ | _🟨 Low–medium_ | _⚠️ Needs verification_

**This assertion still only proves DefaultLight.**

Polling until `body === ::before` is a solid improvement over not-transparent. The note is right that this seeded form has no stored theme, so a hardcoded `#f3f3f3` fallback could still match.

Not asking to block on this — the unit test that prefers `::before` over a different `themeVariables` color already covers the logic. If we have a cheap fixture with a stored theme whose surface is not `#f3f3f3`, one extra poll would lock the actual h930 failure mode. If that is awkward in this suite, calling it out in the PR is enough.

PeterMarinov (commented):
## Review

Nice diagnosis: SurveyJS 3 paints the fill surface on `.sd-root-modern::before`, not via wrapper `themeVariables`. Waiting on `isModelReady` and `onAfterRenderSurvey` is the right contract now that **Hub only supports SurveyJS 3.x** — no need to keep a v2 `--sjs-*` compatibility path.

Inline notes are the items that still look worth a look. A couple need a live check rather than a definite bug. Nits (comment length, CSS last-resort literal, `as CSSProperties`) left out.

**Agent docs:** worth a short `bump-surveyjs` / `DESIGN.md` public-embed bullet so the next bump does not re-guess CSS vars. Not blocking if you prefer a follow-up.

**Merge:** close, after the fallback / restore questions are either confirmed safe or tightened.

PeterMarinov (changes requested):
Right diagnosis, right primary fix (::before + onAfterRenderSurvey + wait for isModelReady). Value is real — embed fill vs SurveyJS 3 cascade. Unit/e2e direction is good.

Not merge-ready as-is. 
1) appliedTheme effect cleanup flashes html/body back to empty;
2) --sjs-* themeVariables fallback reintroduces h930 on the pre-paint path;
3) comment dump; 
4) one short agent-doc bullet so the next bump does not re-guess CSS vars;

e2e/tests/embed/embed-form.spec.ts
orilux: Agreed - this only proves the fallback is unused for `DefaultLight`. I didn't add a themed e2e fixture: there's no cheap seeded form with a stored non-default theme in this suite, and standing one up felt like more surface than this PR should take on.

The mechanism itself (prefer the real `::before` read over any fallback) is now locked at the unit level in `survey-component.test.tsx` - `"uses the survey's own rendered background when already painted"` sets the pseudo-element color to something distinct from `DEFAULT_FILL_BACKGROUND_COLOR` and asserts the rendered color wins. Combined with dropping the `--sjs-*` guesswork per your other comment, the actual h930 failure mode - silently falling back to a wrong/guessed color instead of what SurveyJS rendered - can't reproduce the way it did before. What's left unverified is specifically the real-browser/theme-fixture case, which is a smaller residual risk.

Calling it out here rather than blocking on it - happy to pick up a themed e2e fixture as a fast-follow if one's cheap to seed.

features/public-form/ui/survey-component.tsx
orilux: Removed it; the only fallback left when `.sd-root-modern::before` isn't painted yet is `DEFAULT_FILL_BACKGROUND_COLOR` (the same constant the CSS fallback already uses)

features/public-form/ui/survey-component.tsx
orilux: Split it into two effects: one keyed only on `isFillMode` that captures/restores the pre-fill-mode color, and one that paints/re-paints on `appliedTheme` changes without ever touching the restore. Added a regression test using a `MutationObserver` on `document.body` proving only one background write happens during a theme update (previously would've been two: reset-then-repaint).

PeterMarinov (approved):

PR #941, opened 2026-09-06
fix(h938): Nested loops not getting rendered

### LINKED ISSUE #938: Nested loops are silently failing to render
## Summary

A `paneldynamic` carrying `loopSource` that sits inside another loop's `templateElements` is
invisible to the Question Loops feature. The form loads without error and the designer will
happily let an author configure it, but at runtime the inner loop never produces a single panel.
It fails silently - no console warning, no validation error, no empty state.

## Reproduction

Load this form JSON in the runner (or the editor preview), tick two choices in
`question_outside_of_loop`, then tick two choices in `question_inside_of_loop` inside the first
outer panel:

```json
{
  "pages": [
    {
      "name": "page1",
      "elements": [
        {
          "type": "checkbox",
          "name": "question_outside_of_loop",
          "title": "question_outside_of_loop",
          "choices": [
            { "value": "item1", "text": "outside_Item 1" },
            { "value": "item2", "text": "outside_Item 2" },
            { "value": "item3", "text": "outside_Item 3" }
          ]
        },
        {
          "type": "paneldynamic",
          "name": "question2",
          "title": "parent_loop_panel",
          "templateElements": [
            {
              "type": "checkbox",
              "name": "question_inside_of_loop",
              "title": "question_inside_of_loop",
              "choices": [
                { "value": "item1", "text": "inside_Item 1" },
                { "value": "item2", "text": "inside_Item 2" },
                { "value": "item3", "text": "inside_Item 3" }
              ]
            },
            {
              "type": "paneldynamic",
              "name": "question1",
              "title": "child_loop_panel",
              "templateElements": [{ "type": "rating", "name": "question3" }],
              "loopSource": ["question_inside_of_loop"]
            }
          ],
          "loopSource": ["question_outside_of_loop"]
        }
      ]
    }
  ]
}
```

**Expected:** the outer loop expands to one panel per selected outer choice, and inside each of
those, the child loop expands to one panel per choice selected in *that* panel's
`question_inside_of_loop`.

**Actual:** the outer loop expands correctly (2 panels). The child loop stays at **0 panels**
forever, with `value: undefined`, `isLoopReady: undefined` and `templateVisibleIf: undefined`.

Measured against a `SurveyModel` bound with `bindFeatureToSurvey`. The respondent's answers to
`question_inside_of_loop` *are* stored correctly - only the child loop's panel generation is dead.

A second variant was also measured: pointing the child loop's `loopSource` at a **top-level**
question instead of one inside the outer loop. That fails identically (0 panels). So this is not
specific to the shape above - **no nesting arrangement works today.**

orilux:
## Description

Question Loops did not work inside **any** dynamic panel. A `paneldynamic` carrying `loopSource` that sat inside another panel never produced a single item: the form loaded without error, the designer happily offered the configuration, and the loop was simply dead — no warning, no empty state.

The reported symptom was a loop nested inside another loop. Investigation widened it: two more shapes fail identically, neither involving nesting.

| Arrangement | Before |
|---|---|
| Loop inside a loop (as reported) | 0 panels |
| Loop inside a **plain** `paneldynamic` (container has no `loopSource`) | 0 panels |
| **Top-level** loop sourcing a question that only exists inside some panel template | 0 panels |

One assumption was baked through the whole slice: **a loop panel and its source are both page-level questions addressable by bare name from the survey root.** Six defects followed from it, and each layer is fixed here:

1. **Discovery** — `getAllLoopQuestions` used `survey.getAllQuestions()`, which returns page-level questions only. Replaced with two explicit tree walkers, split by intent: template-level (loop *declarations*, for schema-time visibility injection) and instance-level (live loop *objects*, one per containing panel, for runtime sync). The nesting primitive is any containing `paneldynamic`, not specifically a loop.
2. **Event routing** — sync was bound only to `onValueChanged`, which SurveyJS never raises for a question inside a dynamic panel (it raises it for the container). Added an `onDynamicPanelValueChanged` channel for nested sources, scoped to the panel the change came from.
3. **Source resolution** — `survey.getQuestionByName` cannot see panel-nested questions and could not disambiguate instances if it could. Resolution now goes through the loop's own `data`, the same mechanism SurveyJS carry-forward uses for `choicesFromQuestion`. `panel.<name>` is the canonical stored form; bare names still resolve, innermost scope first, so hand-authored JSON keeps working.
4. **Cascade** — neither event that could carry it fires: `onDynamicPanelAdded` stays silent when panels are created by assigning `value`, which is exactly how they are created. Syncing now descends explicitly, depth-first, inside a per-survey re-entrancy guard (replacing a module-level flag that was shared by every survey on the page).
5. **Exit conditions** — a single-level path (`{inner[0].q}`) silently evaluates to `false` for a nested loop; the real path is `{outer[0].inner[0].q}`. Conditions are now path-aware, and `isLoopExited` resolves its loop by walking up from the evaluating question rather than by name — no lookup, no path parsing, and instances disambiguate for free. The 3-argument signature is unchanged.
6. **Designer** — one source-scope rule for **every** loop, nested or not: siblings in the loop's own template (emitted `panel.<name>`), ancestor and page-level questions (bare), never descendants, never questions from an unrelated panel template.

Loop nesting is capped at depth 2 and enforced at both ends — the runtime does not sync past it and the designer does not offer sources for a loop beyond it.

## Behaviour changes on existing forms

Worth knowing before this ships. None of these are opt-in:

- **Exit conditions now work on single-level loops with prefilled data.** Injection previously only reached panels created *after* it ran, so a model bound to already-loaded data (a resumed submission) kept an uninjected `visibleIf` and exit logic could never hide anything. This was a pre-existing bug, not a nesting one.
- **Exit state is no longer shared between panel instances.** The `exitMeta` property default is a single object, so every loop instance started out holding the *same reference*; triggering an exit in one panel affected its siblings. Invisible at single level — it needs two instances of one declaration to show.
- **The loop source picker offers fewer options.** Sources that could never resolve at runtime are no longer selectable, including for top-level loops. Existing saved values are untouched.

## Notes for reviewers

**`exitMeta` is now `isSerializable: false`.** It is pure runtime state and was round-tripping into `toJSON()` — latent today (design mode never sets it), real the moment a runtime model is serialised.

**One call site changed outside the slice.** `features/submissions/ui/shared/use-survey-model.hook.ts` bound question loops *before* assigning `model.data`. Assigning `survey.data` fires **no events at all** — `setDataCore` writes the values hash and notifies questions directly — so a binding made earlier has nothing to hydrate from, and loops the respondent had not reached stayed empty on reopen. Binding now happens after the assignment, which is the order `features/public-form/` already used. `onModelCreated` keeps its position, so other extensions are unaffected.

**Vendor behaviour is pinned by shape guards.** `__tests__/surveyjs-shape-guards.test.ts` asserts the five SurveyJS behaviours this design depends on, so a `survey-core` upgrade fails there loudly rather than breaking loops silently at runtime. Worth wiring into the `bump-surveyjs` skill checklist as a follow-up.

**Deliberately out of scope:** the h709 `ExtensionModule` migration for this slice (it touches five call sites for reasons unrelated to this fix), and the `bump-surveyjs` checklist entry. A separate defect found while testing — loops still render Add/Remove panel buttons, which destroys answers — is filed separately rather than folded in here.

## Testing

- **234 tests in the slice**, of which **160 are pre-existing and pass unchanged** — single-level loops are in production, so this work had to be provably invisible to them. No existing test was edited.
- Full hub suite: **4440 passing**. The one failure (`no-build-time-client-config`) reproduces on a clean stashed tree and is unrelated to this branch.
- `tsc` error count **identical to baseline** (150, all pre-existing, all in test files); none in changed files. ESLint clean on every changed file.
- New coverage includes the reported JSON end to end, sibling-panel independence, source shrink and grow, a loop inside a plain panel, a page-level source driving a nested loop, resume/hydration for both bind orders, nested exit conditions, the depth cap, and the designer scope rule.

## Related Issues
closes #938 

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [ ] Other (please describe)

## Checklist

> [!NOTE]

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nested question-loop synchronization, including loops inside panels and multiple nesting levels.
  * Corrected loop source resolution to respect panel scope and similarly named questions.
  * Improved restoration of nested loops from prefilled or resumed survey data.
  * Fixed loop exit behavior, visibility updates, and synchronization when source answers change.
  * Prevented synchronization conflicts between sibling loops and separate surveys.
  * Added safeguards for deeply nested loops and improved consistency when loop data changes.


lib/survey-features/question-loops/utils/loop-path.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (!owner) {
+      break;
+    }
+
+    const panel = current.parent as unknown as PanelModel | undefined;
+    const panelIndex = panel ? (owner.panels ?? []).indexOf(panel) : -1;
+    if (panelIndex < 0) {
+      // Still a template element rather than a live instance: no index to bake
+      // in, so the qualified name stops here.
+      break;
```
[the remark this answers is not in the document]
orilux: Addressed

lib/survey-features/question-loops/utils/loop-source-scope.ts
```diff
@@ -0,0 +1,182 @@
...
+  const descendants = getDescendantQuestions(loopQuestion);
+
+  for (const scope of getScopeChain(loopQuestion, survey)) {
+    const match = scope.questions.find(
+      (question) =>
+        question.name === bareName &&
+        question !== loopQuestion &&
+        !descendants.has(question) &&
+        isSelectBaseQuestion(question),
+    );
```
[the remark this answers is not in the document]
orilux: Addressed

lib/survey-features/question-loops/__tests__/surveyjs-shape-guards.test.ts
```diff
@@ -0,0 +1,268 @@
...
+
+const PROBE_FUNCTION_NAMES = ["h938ProbeContext", "h938ProbeIndex"];
+
+/**
+ * Shape guards for the SurveyJS behaviour question-loops nesting depends on (h938).
+ *
+ * These assert vendor behaviour, not our code. They exist so a survey-core upgrade
+ * fails loudly here instead of silently breaking nested loops at runtime.
+ * See `.agents/skills/bump-surveyjs/SKILL.md`.
+ */
```
PeterMarinov: _📝 Documentation_ | _🟡 Medium_

**This file tells agents to update `bump-surveyjs`, but that skill has no loop / shape-guard entry.**

These tests pin vendor behaviour (`getAllQuestions()` misses nested `paneldynamic`, `setDataCore` fires no events, `onDynamicPanelAdded` is silent on `value` assign, condition runner `parentQuestion` walk). A `survey-core` bump that changes any of those will break nested loops the same way h938 did: form loads, designer looks configured, zero panels.

The PR defers the skill update. Without it, the next bump or the h709 `ExtensionModule` move will not know this file exists.

### Proposed fix (keep it short — do not copy the six-defect write-up)

**1.** `hub/.agents/skills/bump-surveyjs/SKILL.md` — add two bullets under Breaking-change checklist:

```markdown
- Question loops (nested `paneldynamic`): run
  `lib/survey-features/question-loops/__tests__/surveyjs-shape-guards.test.ts`.
  A failure here is a vendor-contract break, not a flaky unit test.
- Do not rediscover loops with `survey.getAllQuestions()` (with or without
  `includeNested`). Walk templates vs live instances
  (`collectLoopTemplates` / `collectLoopInstances` in `question-loops/utils/`).
```

**2.** `hub/.agents/skills/add-survey-feature/SKILL.md` — under `onModelReady`, one sentence next to the bind example:

```markdown
Question loops: call `bindFeatureToSurvey` / `bindToSurvey` **after**
`model.data` is assigned. Bulk `setDataCore` raises no `onValueChanged`;
binding first then assigning data leaves nested loops empty on resume
(public-form and submissions `use-survey-model` already use this order).
```

Do not paste the h938 investigation into either skill. Point at this test file + `survey-bindings.ts` (`hydrateLoopsFromData`).

lib/survey-features/question-loops/__tests__/nested-loop-sync.test.ts
```diff
@@ -0,0 +1,325 @@
...
+          itemText: "outside_Item 2",
+          itemValue: "item2",
+          loopIndex: 1,
+          [NESTED_INNER_SOURCE_NAME]: ["item2"],
+        },
+      ],
+    };
+
+    // act
+    bindFeatureToSurvey(survey);
```
PeterMarinov: _🧪 Test Coverage_ | _🟡 Low–medium_ | _⚠️ Needs verification_

**PR claims hydration for both bind orders; this suite only covers data-then-bind.**

`bindFeatureToSurvey` hydrates only if `survey.data` already has keys:

```ts
if (survey.data && Object.keys(survey.data).length > 0) {
  hydrateLoopsFromData(survey);
}
```

Then it subscribes to events. Assigning `model.data` **after** bind uses SurveyJS `setDataCore`: values land, **no** `onValueChanged` / `onDynamicPanelValueChanged`. Nested loops whose panels were not already in that hash stay at 0 items. That is the submissions-viewer bug this PR fixes by moving bind below `model.data = submissionData`.

This test is the **working** order (assign data, then bind). I did not find a sibling that binds first, then assigns `survey.data`, and expects inner panels to stay empty (or to hydrate if you later add that).

Please grep the slice for bind-then-assign. If it is missing, add it next to this case so the hook order cannot regress without a red test.

### Proposed test (agents: put in this same `describe`)

```ts
it("does not rebuild nested panels if data is assigned after bind", () => {
  // Arrange — the order that silently failed in submissions before h938
  const survey = new SurveyModel(nestedLoopSurveySchema as never);
  bindFeatureToSurvey(survey);

  // Act — bulk assign; SurveyJS will not raise value-changed events
  survey.data = {
    [NESTED_OUTER_SOURCE_NAME]: ["item1", "item2"],
    [NESTED_OUTER_LOOP_NAME]: [
      {
        itemText: "outside_Item 1",
        itemValue: "item1",
        loopIndex: 0,
        [NESTED_INNER_SOURCE_NAME]: ["item1", "item3"],
      },
    ],
  };

  // Assert — hydrate never ran; inner loop stays empty
  const outerLoop = survey.getQuestionByName(NESTED_OUTER_LOOP_NAME) as LoopLike;
  // Outer panels may exist from the stored hash even without our sync;
  // the inner loop is what stayed dead. Adjust the assertion to whatever
  // SurveyJS actually materialises — if outer.panels is already 1, still
  // expect inner count 0.
  expect(panelCountOf(outerLoop.panels[0], NESTED_INNER_LOOP_NAME)).toBe(0);
});
```

If that assertion is wrong in jsdom (vendor already expands from `data`), say so on the thread and pin whatever **does** differ from data-then-bind. The point is one test that fails if someone swaps the submissions hook back to bind-then-assign.

Call-site reminder: `features/submissions/ui/shared/use-survey-model.hook.ts` must keep bind **after** `model.data =`. Public-form already does.

PeterMarinov (commented):
## Review

Two follow-ups so this does not silently die on the next SurveyJS bump or the h709 `onModelReady` move. Loop engine itself looks sound.

Inline comments have the proposed edits agents can apply.

PeterMarinov (dismissed):
Approved with 2 docs changes requested as docs drift can cause troubles in the future. Please address before merging

lib/survey-features/question-loops/__tests__/nested-loop-sync.test.ts
orilux: Added test

lib/survey-features/question-loops/__tests__/surveyjs-shape-guards.test.ts
orilux: Addressed

PeterMarinov (dismissed):

PeterMarinov (approved):

PR #942, opened 2026-09-08
fix(h934): Prevent new empty submissions in embed mode

### LINKED ISSUE #934: Embed script creates empty submissions
### Goal

Loading an Endatix form via the embed script must not create a submission. A submission should only be persisted when the respondent interacts with the form or when intentional prefill query parameters are provided.

### Steps to Reproduce

1. Host a page that loads `embed.js` with `data-form-id` set to a published form (no intentional prefill).
2. Open the host page so the iframe loads `/embed/{formId}?parentOrigin=...&embedId=...`.
3. Open Hub → that form → Submissions.

### Actual

Each page visit creates an empty/incomplete submission. Submission metadata includes SurveyJS variables such as `parentOrigin` and `embedId` (and `heightMode` when fill mode is used).

### Expected

No submission is created from embed page-load alone. Reserved embed query params are not treated as prefill variables. Real prefill keys (and `data-prefill`) still apply variables and may create/update a partial.

### Implementation Notes

**RCA:** `embed.js` always appends `parentOrigin` and `embedId` (and optionally `heightMode`) to the iframe URL for postMessage routing. Public-form `useSearchParamsVariables` treats every query param except `token` / `theme` / `language` / `lang` as SurveyJS variables and enqueues a partial submission when any value is new. Those embed handshake/layout params were never in the ignore list, so every genuine embed load looked like “prefill changed” and submitted metadata-only empties.

**Fix direction:** Treat `embedId`, `parentOrigin`, and `heightMode` as reserved public-form query params (same ignore path as `token` / `theme` / language keys). Keep them on the iframe URL for messaging; do not write them as SurveyJS variables or into submission metadata.

Generic iframe shape after embed.js runs:

```html
<iframe
  data-form-id="{formId}"
  src="https://hub.example/embed/{formId}?parentOrigin=https%3A%2F%2Fhost.example&embedId=edxf-{formId}-0-{uuid}"
  allow="clipboard-write"
  loading="lazy"
></iframe>
```

PeterMarinov:
## Summary

Embed page-load no longer creates an empty/incomplete submission. Handshake query params (`embedId`, `parentOrigin`, `heightMode`) are reserved, not treated as SurveyJS prefill.

Fixes #934.

## Behavior

- **Before:** `embed.js` always appends those keys to the iframe URL. Public-form prefill treated every unknown query param as a survey variable and enqueued a metadata-only partial on first init.
- **After:** reserved keys are ignored by prefill. No submission on embed load. Intentional prefill (`data-prefill` / extra query keys) still applies variables and may create/update a partial.
- **Unchanged:** keys stay on the iframe URL for postMessage routing and fill-height layout. `token` / `theme` / `language` / `lang` stay ignored.

No new Hub/API endpoints. No new client events. Embed custom events (`endatix:form-loaded`, `endatix:form-complete`, `endatix:resize`, …) unchanged.

## Key decisions

- One shared, framework-free module: `features/embed-form/embed-query-params.ts` — imported by `src/embed` (relative) and by public-form (alias).
- Prefill vs handshake: `IGNORED_PARAMS` includes `EMBED_RESERVED_QUERY_PARAMS`. `cleanupUrl` strips prefill only, so messaging can still read the iframe query string.
- ESLint `no-restricted-imports` on `src/embed/**` and `embed-query-params.ts` (no `next` / `react` / `@/*`) so the esbuild IIFE cannot pick up the app.

## Compatibility

- Not a breaking API change.
- Hosts using only the embed script: stop seeing empty rows per page visit.
- Hosts using real prefill beside embed params: still works; reserved keys are not written into submission metadata.


features/embed-form/embed-query-params.ts
```diff
@@ -0,0 +1,10 @@
+/** Iframe query keys for postMessage/layout. Framework-free: `src/embed` and the Next app both import this. Not public-form prefill. */
+export const EMBED_ID_QUERY_PARAM = "embedId";
+export const PARENT_ORIGIN_QUERY_PARAM = "parentOrigin";
+export const HEIGHT_MODE_QUERY_PARAM = "heightMode";
+
+export const EMBED_RESERVED_QUERY_PARAMS = [
```
orilux: `EMBED_RESERVED_QUERY_PARAMS` is a second handwritten copy of the three named exports. `embed.ts` can start writing a new handshake key via a new constant without pushing it onto this array — that is exactly how h934 shipped (`embedId` / `parentOrigin` / `heightMode` existed on the iframe URL but not in `IGNORED_PARAMS`).

Please make membership automatic (single object/`Object.values`, or derive the ignore list from the same record `embed.ts` uses to `.set()`), and add an embed SDK test: every query key the SDK adds, other than `token` and `data-prefill`, must be in `EMBED_RESERVED_QUERY_PARAMS`. The new hook tests re-spell `"embedId"` / `"parentOrigin"` / `"heightMode"` and would stay green through that regression.

features/public-form/application/__tests__/use-search-params-variables.hook.test.ts
orilux: These tests hardcode `"embedId"`, `"parentOrigin"`, and `"heightMode"` instead of iterating `EMBED_RESERVED_QUERY_PARAMS`. That fights the AGENTS.md rule this PR added (“Do not re-spell the names; this is one module”).

A fourth reserved key added to the shared array would be ignored in production and untested here. Drive the arrange/assert loops from the exported list (still pin the public strings in `src/embed/embed.test.ts` if you want a contract test on the wire names).

eslint.config.mjs
```diff
@@ -88,6 +88,24 @@ const eslintConfig = defineConfig([
       ],
     },
   },
+  {
+    files: ["src/embed/**/*.ts", "features/embed-form/embed-query-params.ts"],
+    ignores: ["src/embed/__tests__/**"],
```
orilux: The restriction applies to `src/embed/**/*.ts`, but the ignore is only `src/embed/__tests__/**`. The iframe URL tests live in `src/embed/embed.test.ts`, so this ignore does not cover them.

Either ignore `src/embed/**/*.test.ts` (if tests are allowed to import `@/`), or drop the ignore if the IIFE constraint should apply to tests too. As written it looks like it exempts embed tests and does not.

orilux (dismissed):

features/embed-form/embed-query-params.ts
PeterMarinov: Agreed: array was a second copy left unnotices. Now, single EMBED_QUERY_PARAMS record; named exports + EMBED_RESERVED_QUERY_PARAMS = Object.values(...). SDK test in src/embed/embed.test.ts: iframe query keys besides token must be in that list (heightMode=fill + token).

features/public-form/application/__tests__/use-search-params-variables.hook.test.ts
PeterMarinov: Agreed. Arrange/assert loops EMBED_RESERVED_QUERY_PARAMS (ignore, mixed prefill, cleanup keep, reserved-only no replace). Wire names still pinned in the embed SDK test.

eslint.config.mjs
PeterMarinov: Agreed: ignore was only src/embed/__tests__/**, so src/embed/embed.test.ts was still under the IIFE import ban. Now src/embed/**/*.test.ts plus __tests__. Tests may import @/; production embed IIFE still cannot.

orilux (approved):

PR #953, opened 2026-09-10
fix(h939): Switching to Preview tab drops changes

### LINKED ISSUE #939: In dev mode the form builder drops unsaved questions when switching to Preview
### Goal

Keep unsaved Form Builder canvas changes when switching between Design and Preview. Preview must not reload the last saved definition over in-memory edits.

### Steps to Reproduce

1. Open Form Builder on `localhost` (`next dev`), e.g. `/forms/{formId}/design`.
2. Add a question to the canvas. Do **not** save.
3. Click the Preview tab.
4. Wait about half a second, then switch back to Design.

Also compare: save first, then Preview — the question should survive. Production on the same branch does not reproduce this.

### Actual

Preview shows the new question briefly (~500ms), then the question disappears. Returning to Design, the question is gone. Unsaved work is lost.

### Expected

The unsaved question stays in Preview and is still on the canvas after returning to Design. Tab changes must not reload persisted JSON over live Creator state.

### Implementation Notes

Two cooperating pieces:

1. Preview writes `?tab=preview` with `router.replace` (`useCreatorTabUrl` → `useUrlSearchParamsUpdater`). That is a real App Router navigation, not a shallow history update. `startTransition` keeps the current Preview visible until the server round-trip finishes (the flash).
2. `next dev` always re-renders the design Server Component (pages are never cached in development). `design/page.tsx` refetches the definition and `JSON.parse` yields a **new** `formJson` object. `FormEditor` then does `creator.JSON = formJson` on any new reference, with no unsaved-changes guard — overwriting the live survey with the last saved definition.

The design page does not read `searchParams`; the refetch exists only because the tab was written with `router.replace`. Production does not disable that cache the same way, so `formJson` is not replaced while editing.

The same `creator.JSON = …` pattern exists on the template editor.

**Likely fix (do one or both):**

- Set Creator JSON once on init (or skip the assignment while there are unsaved changes).
- Write `?tab=` with `window.history.replaceState` instead of `router.replace` (Next.js supported shallow URL sync; `useSearchParams` still updates).

**Confirm in DevTools:** add a question, do not save, click Preview. URL becomes `?tab=preview`. Network shows an RSC request for the design URL; the question vanishes when it completes.

orilux:
## Summary

In `next dev`, switching Design → Preview dropped unsaved canvas questions. Preview wrote `?tab=preview` with `router.replace`, which is a real App Router navigation. The design Server Component refetched, `JSON.parse` produced a new `formJson` object, and `creator.JSON = formJson` overwrote live Creator state with the last saved definition.

Production did not reproduce this because pages are cached; `next dev` always re-renders.

## What changed

Two cooperating fixes, both needed:

1. **Shallow `?tab=` writes** — `useCreatorTabUrl` now uses `history.replaceState`. `useSearchParams` still updates (Next.js patches the History API); the Server Component does not refetch.
2. **Content-aware JSON apply** — `useCreatorJson` assigns `creator.JSON` when the snapshot *content* changes, not when the object identity changes. A Server Component refetch of the same definition is skipped; a distinct snapshot (AI chat turn) is still applied. The template editor uses the same hook.

`updateFormDefinitionJsonAction` / `updateTemplateJsonAction` also `revalidatePath` on save, which is a second source of a new `formJson` reference on the same record. The content guard covers that path too.

## Related Issues

Fixes #939

## Test plan

- [ ] Open Form Builder on `localhost` (`next dev`), e.g. `/forms/{formId}/design`
- [ ] Add a question. Do **not** save
- [ ] Switch to Preview, wait ~500ms, switch back to Design — the question is still there
- [ ] Confirm DevTools: URL becomes `?tab=preview`, but there is **no** RSC request for the design URL
- [ ] Save first, then Preview — still works (existing path)
- [ ] With AI assistant enabled, run two chat turns that edit the form — canvas reflects turn 2, not frozen at turn 1
- [ ] Repeat unsaved Design ↔ Preview on a form template

## Notes for reviewers

- Table/list URL updates still go through `useUrlSearchParamsUpdater` → `router.replace`. That must keep navigating so paged lists refetch. Do not fold Creator tab writes into that hook.
- `useCreatorJson` is **not** apply-once-per-instance. That would freeze AI chat after the first turn (`form-editor-with-chat` pushes each `resultDefinition` into the same Creator).
- JSON apply runs after the `onModified` listener so a bulk load is filtered as `JSON_CHANGED_TYPE`, matching the previous effect order.

## Type of Change
Check all that apply.

- [ ] Documentation update
- [ ] Dependency update
- [ ] Other (please describe)

## Checklist

> [!NOTE]

## Summary by CodeRabbit

* **Bug Fixes**
  * Survey Creator edits are preserved when the same form data is refreshed.
  * New or updated form data is applied correctly, including when switching editor instances.
  * Switching between Designer and Preview tabs updates the URL without triggering unnecessary page refreshes.

* **Tests**
  * Added coverage for form data synchronization and editor tab URL updates.
  * Added validation for search-parameter URL changes.


features/forms/ui/editor/form-editor.tsx
```diff
@@ -599,6 +602,8 @@ function FormEditor({
     return () => creator.onModified.remove(setAsModified);
   }, [creator, setHasUnsavedChanges]);
 
+  useCreatorJson(creator, formJson);
```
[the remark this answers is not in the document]
orilux: This is a bugfix PR. Refactoring major feature is not part of the objectives.

lib/survey-features/survey-design/use-cases/should-apply-creator-json.ts
```diff
@@ -0,0 +1,33 @@
...
+function isSameSurveyJson(left: object | null, right: object): boolean {
+  if (left === right) {
+    return true;
+  }
+
+  if (!left) {
+    return false;
+  }
+
+  return JSON.stringify(left) === JSON.stringify(right);
```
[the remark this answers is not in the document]
orilux: Addressed

lib/survey-features/survey-design/use-cases/should-apply-creator-json.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  if (Array.isArray(left) && Array.isArray(right)) {
+    if (left.length !== right.length) {
+      return false;
+    }
+
+    return left.every((item, index) => isSameSurveyJson(item, right[index]));
+  }
+
```
PeterMarinov: It's better to use SurveyJs's own `Helpers.isTwoValueEquals` to compare survey models instead of duplicating the logic. These methods are also used internally, so our change detection will be exactly the same as SJS if we stick to the helpers

lib/survey-features/survey-design/ui/use-creator-tab-url.ts
```diff
@@ -1,18 +1,35 @@
...
+ */
+function replaceTabQuery(nextQueryValue: string | null) {
+  const href = hrefWithSearchParamUpdates(globalThis.window.location.href, {
+    [CREATOR_TAB_QUERY_KEY]: nextQueryValue,
+  });
+  if (!href) {
+    return;
+  }
+
+  globalThis.window.history.replaceState(null, "", href);
```
PeterMarinov: This causes Themes tab to detect changes as it re-assigns the theme on every tab visit asking users to "Save detected theme changes"

lib/survey-features/survey-design/use-cases/should-apply-creator-json.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (appliedTo !== creator) {
+    return true;
+  }
+
+  return !isSameSurveyJson(lastAppliedJson, json);
+}
+
+function isSameSurveyJson(left: unknown, right: unknown): boolean {
```
PeterMarinov: **Fix (local `6ed74b0b`):** delete this file.

Hand-rolled deep compare duplicates SurveyJS and still misses definition vs answer semantics. Use `Helpers.checkIfValuesEqual` from `survey-core` **inside** `useCreatorJson`, with:

```ts
{
  ignoreOrder: false,
  caseSensitive: true,
  trimStrings: false,
  doNotConvertNumbers: true,
}
```

`isTwoValueEquals` is the wrong helper here — it cannot pass `doNotConvertNumbers`, so a `rateMax: "5"` vs `5` edit would be treated as the same snapshot and skipped.

Also compare against `creator.JSON`, not only `lastAppliedJson`: after save, `revalidatePath` returns content that already sits on the canvas, and assigning still wipes undo/selection.

Regression tests live on `use-creator-json.test.ts` (key order, array order, case-only title, string→number).

lib/survey-features/survey-design/ui/use-creator-json.ts
```diff
@@ -0,0 +1,39 @@
...
+        creator,
+        json,
+        appliedToRef.current,
+        lastAppliedJsonRef.current,
+      )
+    ) {
+      return;
+    }
+
+    creator.JSON = json;
```
PeterMarinov: **Keep this hook as the only apply path** (form + template editors already call it).

Remote version only skips when `lastAppliedJson` matches. Local commit also skips when `creator.JSON` already equals the incoming snapshot — that is the save/`revalidatePath` path. Assigning `creator.JSON` is not free: it resets undo and selection.

Do not make this apply-once-per-instance: AI chat pushes a new `resultDefinition` onto the same Creator.

lib/survey-features/survey-design/ui/use-creator-tab-url.ts
```diff
@@ -1,18 +1,35 @@
...
+}
```
PeterMarinov: **Keep `history.replaceState`.** That is the h939 navigation fix (`router.replace` → RSC refetch → clobber canvas). Lists must keep `useUrlSearchParamsUpdater` so they still refetch.

The Themes “Save detected theme changes” false positive is **not** this History write. `ThemeTabPlugin.activate` and reapplying `availableThemes` raise the same events as a user edit. Local `ae3cb786` wraps those in `hydrateThemeTab` in `use-theme-management.hook.ts` and only clears `hasPendingThemeChanges` when Hub `isThemeDirty` is false.

Local `8e2a9925` also inlines this one-key URL write and drops `href-with-search-param-updates.ts` — a generic helper for a single query key was extra surface.

lib/survey-features/survey-design/use-cases/href-with-search-param-updates.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Same-origin path+search+hash after applying search-param updates, or `null`
+ * when the query would not change. Empty / null values delete the key.
+ */
+export function hrefWithSearchParamUpdates(
+  currentHref: string,
+  updates: Record<string, string | null>,
+): string | null {
```
PeterMarinov: **Fix (local `8e2a9925`):** delete this helper + its tests. `replaceTabQuery` can `URLSearchParams.set`/`delete` one key in place. No second public API for the same job.

AGENTS.md
```diff
@@ -14,7 +14,8 @@
 Prefer vendor types from `survey-core` / `survey-creator-core`. Where they widen to `string` (Creator `activeTab`), the closed Hub union lives in [`lib/survey-js/`](lib/survey-js/) — built-in ids plus Hub plugin ids. Import it instead of re-spelling those strings in a slice (`FORM_DIAGNOSTICS_PLUGIN_NAME` is `ENDATIX_CREATOR_TAB.diagnostics`).
 
 - Preview's id is `preview`; `test` is a legacy id Creator still emits — normalise with `canonicalizeCreatorTabId`, never compare raw.
-- `?tab=` slugs are one map in `creator/tab-url.ts` (`design` → `designer`; Design omits the param).
+- `?tab=` slugs are one map in `creator/tab-url.ts` (`design` → `designer`; Design omits the param). Write `?tab=` with `history.replaceState`, never `router.replace` — App Router navigation refetches the saved definition over live Creator state (h939).
+- Apply Creator JSON when its content changes (`useCreatorJson`). Skip a new object that is the same snapshot (dev RSC refetch). Do not skip a distinct snapshot — AI chat turns replace the canvas on the same Creator.
```
PeterMarinov: **Tighten these two bullets** (already done locally). The remote text is directionally right but incomplete:

- Tab writes: `history.replaceState` in `useCreatorTabUrl`, **and** paged lists still use `useUrlSearchParamsUpdater` (do not fold them together).
- JSON: never assign `creator.JSON` in a slice — `useCreatorJson` only; skip if already applied **or already on the canvas**.
- Compare with `Helpers.checkIfValuesEqual` + all four flags; do not document `isTwoValueEquals` for definitions.
- Theme hydration: `hydrateThemeTab`; Hub dirty vs Creator `hasPendingThemeChanges`.

Keep it to those four bullets. The trap (answer-comparison defaults) is worth one sentence so the next agent does not “simplify” the flags away.

PeterMarinov (commented):
## Follow-up on the remote diff vs three local commits

Remote head (`0c9be65`) already has the right **two-layer** h939 fix:

1. `history.replaceState` for `?tab=` so App Router does not refetch saved JSON over live Creator state.
2. Guarded `creator.JSON` apply so a new object with the same snapshot is skipped, while a distinct snapshot (AI chat) still applies.

Three **local, unpushed** commits on this branch (`ae3cb786`, `8e2a9925`, `6ed74b0b`) address the remaining review findings and simplify the code. Inline comments below are on **this** head so they map to the current Files tab; after push some of these files are deleted on purpose.

### What those commits change (and why)

- **Do not hand-roll survey equality.** Drop `should-apply-creator-json.ts`. Compare with `Helpers.checkIfValuesEqual` and **all four flags** (`ignoreOrder: false`, `caseSensitive: true`, `trimStrings: false`, `doNotConvertNumbers: true`). Defaults are *answer* comparison (case-fold, trim, `"5"` → `5`). `isTwoValueEquals` cannot pass `doNotConvertNumbers` — so that earlier suggestion is close but not enough. Also skip when `creator.JSON` already matches (save → `revalidatePath`).
- **Theme dirty false-positive is not `replaceState`.** `ThemeTabPlugin.activate` / chooser `availableThemes` fire the same events as a user edit. Wrap those in `hydrateThemeTab`; clear `hasPendingThemeChanges` only when Hub `isThemeDirty` is false. Keep `replaceState` for tabs.
- **Collapse ceremony.** Inline the one-key URL write; delete `href-with-search-param-updates`. Keep `useCreatorTabUrl` / `useCreatorJson` as the only public seams.
- **Agent docs.** Keep four short SurveyJS bullets (History vs list `router.replace`, never assign `creator.JSON` in a slice, `checkIfValuesEqual` + flags, `hydrateThemeTab`). No extra essays.

### Tests (right layer; skip a Playwright Creator e2e here)

Unit coverage that actually pins the bugs: refetch vs AI turn vs save-revalidate; key-order / case / `"5"` vs `5`; `replaceState` + hash/other params + no-op; Theme Editor activate / chooser / mid-edit reopen. An integration e2e of full Creator would be brittle and slow for this.

### Out of scope (agree with existing thread)

Do not move `form-editor.tsx` into a verb-noun slice in this bugfix.

`features/forms/ui/preview/preview-form.tsx` still assigns `creator.JSON` on a **new** Creator — not the live Design canvas. Leave it.

Push the three local commits when ready so these threads resolve against the real code.

PeterMarinov (changes requested):
Great to extract the use-creator-json for single place to mutate the Creator instance json, but the PR has also introduced a bug with change detection on opening of the Themes tab + on Save changes. 

See the comments, I will address them shortly

PeterMarinov (dismissed):
All comments seem to be addressed. Let's do final test and merge

PeterMarinov (approved):

PR #989, opened 2026-09-16
feat(h988): export Hub OpenTelemetry logs on every host

### LINKED ISSUE #988: Export Hub OpenTelemetry logs on every host
### Goal
Make Hub app logs first-class OpenTelemetry signals so operators see them in Azure Application Insights (`traces`), an OTLP collector (AWS / self-hosted), and host stdout — not only HTTP/Next.js spans.

Today spans work when `APPLICATIONINSIGHTS_CONNECTION_STRING` is set at runtime, but `TelemetryLogger` records do not appear in Kusto `traces` / `exceptions`. Console fallback is also muted once an exporter is configured, so SWA/K8s log streams stay empty.

### Implementation Notes
- One `NodeSDK` for all hosts: OTLP exporters when `OTEL_EXPORTER_OTLP_ENDPOINT` is set, Azure Monitor trace+log exporters when the App Insights connection string is set (fan-out if both). Do not run a second SDK (`useAzureMonitor` returning null skipped flush).
- Keep Next.js `instrumentation.ts` + Node-only `instrumentation.node.ts`. Mark OTel / Azure exporter packages as `serverExternalPackages` so `@opentelemetry/api-logs` is a process singleton.
- Always flush on shutdown. Stdout structured logs via the Logs pipeline (12-factor).
- App coverage: PDF render timeout is a warn log (expected), not an exception.
- Docs: Hub observability (env vars, SWA runtime vs build, disable classic App Insights site agent, which App Insights tables to query).
- Do not put the App Insights connection string in the SWA **build** environment; runtime app settings only.

PeterMarinov:
## Summary
Closes #988. Hub already exported **spans** (App Insights `requests` / `dependencies`); `TelemetryLogger` records never reached the OTel Logs API (`traces` stayed empty). One `NodeSDK` strategy, explicit `LoggerProvider` (same split as `@vercel/otel`), Azure Monitor **exporters** instead of `useAzureMonitor`.

Hub does **not** use `@vercel/otel` (traces-first; Vercel drain). Start stays Next `instrumentation.ts` + Node-only `instrumentation.node.ts`.

### Removed
- `AzureTelemetryStrategy` — distro returned `null`, skipped flush, Logs API never bound
- `OtelTelemetryStrategy` — second SDK path
- unused `lib/otel-logger.ts`

### One strategy, many exporters
`NodeSdkTelemetryStrategy` fans out traces and logs when both Azure and OTLP are set.

| Destination | Env | Notes |
| --- | --- | --- |
| Azure Application Insights | `APPLICATIONINSIGHTS_CONNECTION_STRING` | **Runtime** app setting only (not SWA build). Logs → `traces`; `error`/`critical` with an `Error` → `exceptions`. No Live Metrics / perf counters. Drop classic `ApplicationInsightsAgent_EXTENSION_VERSION`. |
| OTLP | `OTEL_EXPORTER_OTLP_ENDPOINT` | Default protocol **`grpc`** (port 4317), matching the API — not the spec default `http/protobuf`. |
| OTLP HTTP | `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` or `http/json` | Usually port 4318. Per-signal: `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` / `_LOGS_PROTOCOL`. Unknown protocol → warn + gRPC. |
| Per-signal OTLP | `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` / `_LOGS_ENDPOINT` | Turns on that signal only. Traces-only does **not** count as an active **log** exporter — console fallback still applies. |
| Auth / TLS | `OTEL_EXPORTER_OTLP_HEADERS`, `_CERTIFICATE`, `_CLIENT_KEY` / `_CLIENT_CERTIFICATE`, `_INSECURE` | Exporters read the standard OTel env vars (gRPC metadata or HTTP headers). |
| JSON stdout | `TELEMETRY_CONSOLE_FALLBACK=true` **and** an active log exporter | One JSON object per line (`timestamp`, `severity`, `logger`, `body`, `traceId`, `spanId`, `attributes`). Without a log exporter, the same flag uses the formatted console fallback. |
| Off | no exporter, or `OTEL_SDK_DISABLED=true` | SDK does not start. |

Also: `UndiciInstrumentation` (not `FetchInstrumentation`) so Hub → API `fetch` carries `traceparent`; `serverExternalPackages` so Next does not duplicate `@opentelemetry/api-logs`; default logger name `endatix-hub`; PDF render timeout → `TelemetryLogger.warn` (not `exceptions`); credential-like attribute keys redacted to `[REDACTED]` on every destination.

Docs: [Hub observability](https://docs.endatix.com/docs/developers/hub/observability) (`oss/docs/endatix-docs/docs/developers/hub/observability.mdx`) — protocols, per-signal endpoints, headers/TLS, stdout vs fallback, shutdown (flush, not own SIGTERM).

## Test plan
- [ ] `pnpm test run features/telemetry`
- [ ] Runtime App Insights: `TelemetryLogger` rows in `traces` (`cloud_RoleName == "endatix-hub"`)
- [ ] OTLP gRPC / `http/protobuf` / `http/json` and per-signal endpoints
- [ ] SWA / kubectl log stream: JSON lines when `TELEMETRY_CONSOLE_FALLBACK=true` with a log exporter
- [ ] PDF timeout: warn log, no exception row

## Summary by CodeRabbit

- **New Features**
  - Added OpenTelemetry configuration options for service naming, resource attributes, protocols, headers, signal endpoints, and SDK disablement.
  - Added support for Azure Application Insights and OTLP trace/log exporters, including configurable console fallback.
  - Telemetry output now uses JSON lines and redacts credential-like attributes.

- **Bug Fixes**
  - PDF export timeouts and failures now produce clearer warning and error telemetry.
  - Improved telemetry shutdown and signal handling to flush reliably.

- **Documentation**
  - Expanded telemetry setup, environment variable, logging, and production configuration guidance.


PeterMarinov:
@coderabbit, review

features/pdf-export/submission/render-submission-pdf.use-case.tsx
```diff
@@ -136,13 +134,45 @@ export async function renderSubmissionPdf({
     );
   } catch (error) {
     if (isPdfRenderTimeout(error)) {
-      return Result.error(
-        "PDF render exceeded the deadline.",
-        undefined,
-        PDF_RENDER_TIMEOUT_CODE,
-      );
+      return deadlineExceeded(caller);
```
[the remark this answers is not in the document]
PeterMarinov: Fixed: prepare-model now catches `PDF_RENDER_TIMEOUT_CODE` inside the span callback, sets `pdf.outcome=timeout`, and returns without throwing — so `traceAsync` never `recordException` / error status. Test asserts that.

features/telemetry/__tests__/node-sdk-telemetry-strategy.test.ts
```diff
@@ -0,0 +1,289 @@
...
+      expect(strategy.name).toBe("none");
+    });
+  });
+
+  it("writes JSON stdout only when TELEMETRY_CONSOLE_FALLBACK is true", () => {
+    vi.stubEnv("OTEL_EXPORTER_OTLP_ENDPOINT", OTLP_GRPC);
+    const quiet = initialize().strategy;
+
+    vi.stubEnv("TELEMETRY_CONSOLE_FALLBACK", "true");
+    const forced = initialize().strategy;
```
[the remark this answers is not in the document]
PeterMarinov: Done: `logs.disable()` in `afterEach` so a later `initialize()` can register a new global provider. The JSON-stdout assertion still checks the processor list on the strategy that was just initialized (the exporter wiring under test). Emitting through `logs.getLogger` after two inits would still hit the first registered provider unless we disable — which we now do between tests.

features/telemetry/__tests__/otel-server-externals.test.ts
```diff
@@ -0,0 +1,53 @@
+import { describe, it, expect } from "vitest";
+import { readdirSync, readFileSync } from "node:fs";
+import path from "node:path";
+import { OTEL_SERVER_EXTERNAL_PACKAGES } from "../infrastructure/otel-server-externals";
+
+const INFRASTRUCTURE_DIR = path.resolve(__dirname, "../infrastructure");
+const TELEMETRY_IMPORT = /from\s+"((?:@opentelemetry|@azure)\/[^"]+)"/g;
```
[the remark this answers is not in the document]
PeterMarinov: Done: import regex accepts `'"` quotes. `next.config.ts` assertion now requires `serverExternalPackages` before `OTEL_SERVER_EXTERNAL_PACKAGES` in the same file (not a full TS AST — Hub tests stay regex, enough to catch a stray mention).

features/telemetry/infrastructure/json-console-log-record-exporter.ts
```diff
@@ -0,0 +1,80 @@
...
+import type {
+  LogRecordExporter,
+  ReadableLogRecord,
+} from "@opentelemetry/sdk-logs";
+import { redactSensitiveAttributes } from "./redact-sensitive-attributes";
+
+type LineWriter = (line: string) => void;
+
+const writeToStdout: LineWriter = (line) => {
+  process.stdout.write(line);
```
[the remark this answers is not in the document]
PeterMarinov: Won't do drain/`forceFlush` wait on stdout. This exporter is a debug/ops JSON line to `process.stdout`; a full backpressure queue is disproportionate vs `try/catch` on sync `write`. EPIPE already fails the export callback. Unhandled `error` on stdout is a process-level concern, not this processor.

features/telemetry/infrastructure/strategies/node-sdk-telemetry-strategy.ts
```diff
@@ -0,0 +1,249 @@
...
+        }),
+      }),
+    );
+  }
+
+  if (plan.otlpLogsProtocol) {
+    processors.push(
+      new BatchLogRecordProcessor({
+        exporter: OTLP_LOG_EXPORTERS[plan.otlpLogsProtocol](),
+      }),
```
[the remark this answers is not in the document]
PeterMarinov: Fixed: `hasActiveLogExporter()` is Azure or OTLP **logs** only. Traces-only OTLP still starts the SDK for spans, but `TelemetryLogger` keeps the console fallback. Also redacting attributes **before** `logger.emit` so Azure/OTLP get `[REDACTED]` too.

features/telemetry/infrastructure/telemetry-initializer.ts
```diff
@@ -1,98 +1,151 @@
...
-          .then(
-            () => console.log("Telemetry SDK shut down successfully"),
-            (err) => console.error("Error shutting down Telemetry SDK", err),
-          )
-          .finally(() => process.exit(0));
-      }
+    const flush = () => {
+      strategy
+        .forceFlush()
+        .catch((error) => console.error("Error flushing telemetry", error));
```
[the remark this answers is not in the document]
PeterMarinov: Won't own Next shutdown. Next 16 already registers SIGTERM/SIGINT, drains requests, then `process.exit`. Hub must only **start** `forceFlush` and leave pipelines open for in-flight work — awaiting flush then exiting would skip Next's drain (and `NEXT_MANUAL_SIG_HANDLE` would require Hub to own the whole server, which we explicitly do not). Documented in `telemetry-initializer` and observability.mdx Shutdown.

features/telemetry/infrastructure/telemetry-logger.ts
```diff
@@ -271,33 +250,44 @@ export class TelemetryLogger {
...
+   * and the error message is appended to the body.
+   */
+  private static logFailure(
+    severity: LogSeverity.Error | LogSeverity.Critical,
+    message: string,
+    error: unknown,
+    attributes: LogAttributes = {},
+    loggerName?: string,
+  ): void {
+    if (!error) {
```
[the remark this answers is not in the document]
PeterMarinov: Fixed: `error === undefined` only. `null` / `0` / `false` / `""` now go through `parseErrorMessage` and exception attributes. Test for `null`.

README.md
```diff
@@ -123,7 +123,7 @@ This is useful for testing the production build locally e.g. test with enabled c
 
 Use `TelemetryLogger` from `features/telemetry` for application logs instead of calling `console` directly. The logger emits OpenTelemetry log records when Azure Application Insights (`APPLICATIONINSIGHTS_CONNECTION_STRING`) or OTLP (`OTEL_EXPORTER_OTLP_ENDPOINT`) is configured.
 
-When no telemetry exporter is configured, `TelemetryLogger` mirrors logs to the console in local development so failures remain visible. Production console fallback is off by default to avoid duplicate logs; enable it with `TELEMETRY_CONSOLE_FALLBACK=true` only when the host intentionally collects stdout/stderr.
+When no telemetry exporter is active, `TelemetryLogger` mirrors logs to the console in local development so failures remain visible. Production stdout is off by default to avoid duplicate logs; enable it with `TELEMETRY_CONSOLE_FALLBACK=true` only when the host intentionally collects stdout/stderr. With an exporter configured, that flag writes one JSON object per line.
```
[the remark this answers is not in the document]
PeterMarinov: Fixed: README now says **active log exporter**. `OTEL_SDK_DISABLED=true` still uses formatted console fallback when `TELEMETRY_CONSOLE_FALLBACK=true`, not the JSON processor (SDK never starts).

features/telemetry/__tests__/telemetry-sdk.test.ts
```diff
@@ -0,0 +1,289 @@
...
+      expect(strategy.name).toBe("none");
+    });
+  });
+
+  it("writes JSON stdout only when TELEMETRY_CONSOLE_FALLBACK is true", () => {
+    vi.stubEnv("OTEL_EXPORTER_OTLP_ENDPOINT", OTLP_GRPC);
+    const quiet = initialize().strategy;
+
+    vi.stubEnv("TELEMETRY_CONSOLE_FALLBACK", "true");
+    const forced = initialize().strategy;
```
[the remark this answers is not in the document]
PeterMarinov: Fixed: after `logs.disable()`, a second `initialize()` with JSON stdout is now asserted by emitting through `logs.getLogger` (`json-stdout-canary`) and checking `process.stdout.write` — not only the processor list on the instance.

features/telemetry/infrastructure/json-console-log-record-exporter.ts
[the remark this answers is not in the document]
PeterMarinov: Fixed: stdout writes wait on the Node write callback; `export` chains those promises; `forceFlush`/`shutdown` await the chain so success is not reported before the line is flushed. Tests cover a still-pending write. Not a full drain-queue/backpressure buffer — sequential await of write callbacks is enough for this JSON-line exporter.

igdragoev (commented):
Review of `cdddce13`. There are inline comments on 14 issues. The two most urgent are the failing Build & Test check (`string-utils.test.ts`) and the noise filter no longer matching HTTP server spans while Azure sampling is gone. Together those would sharply increase App Insights volume.

Several comments are about silent telemetry loss (no console fallback when the SDK fails to start, one bad exporter stopping every destination, gRPC TLS default, flush before drain) or behaviour changes on Azure (crash handlers exit the process, duplicate fetch dependencies).

Smaller items, no inline comment:
- `stringifyUnknown` has four options nothing uses.
- The timeout branch in the PDF export's outer catch can no longer be reached.
- Every log call re-reads about 8 env vars. Consider reading config once.
- The description says traces-only OTLP with `TELEMETRY_CONSOLE_FALLBACK=true` gives formatted console output, but the code prints JSON lines.

features/telemetry/infrastructure/filtering-span-processor.ts
```diff
@@ -37,71 +31,41 @@ const FILTER_PATTERNS: readonly string[] = [
...
 ];
 
-function getSpanUrlOrName(span: ReadableSpan): string {
-  const url =
+function shouldDrop(span: ReadableSpan): boolean {
+  if (span.kind === SpanKind.INTERNAL && span.name.startsWith("metric.")) {
+    return true;
+  }
+  const urlOrName = `${
     span.attributes["http.url"] ??
```
igdragoev: **The noise filter no longer drops HTTP server spans. Azure sampling is also gone, so everything gets exported.**

`shouldDrop` looks at `http.url` / `url.full` / `http.target` and the span name. `@opentelemetry/instrumentation-http` 0.221 only emits the stable semantic conventions: the span name is just `GET` and the path is in `url.path` / `url.query`. None of the checks match. I confirmed this with the real `NodeSDK` + `HttpInstrumentation`: `GET /_next/static/chunks/app.js`, `/forms?_rsc=abc` and `/api/health` were all exported. The old `@azure/monitor-opentelemetry` pulled in instrumentation-http 0.219, which still set `http.target`, so this is a regression for Azure. It now also affects OTLP.

On top of that, replacing `useAzureMonitor` with raw exporters (`telemetry-sdk.ts:177`) removed the package's default `RateLimitedSampler` (`tracesPerSecond: 5` in 1.18.2). `AlwaysOnSampler` now sends 100% of traces.

Together: App Insights gets a `requests` row for every static asset, RSC navigation and health probe, with no sampling cap. Ingestion volume and cost will rise sharply, and no setting limits it.

Suggestion: match on `url.path` (keep the legacy keys as a fallback). Restore a rate-limited or ratio sampler for Azure, ideally configurable.

lib/utils/__tests__/string-utils.test.ts
```diff
@@ -15,3 +15,72 @@ describe("stripTrailingNewlines", () => {
...
+
+describe("stringifyUnknown", () => {
+  it("renders nullish and scalars without [object Object]", () => {
+    // Act & Assert
+    expect(stringifyUnknown(null)).toBe("null");
+    expect(stringifyUnknown(undefined)).toBe("undefined");
+    expect(stringifyUnknown("hello")).toBe("hello");
+    expect(stringifyUnknown(0)).toBe("0");
+    expect(stringifyUnknown(false)).toBe("false");
+    expect(stringifyUnknown(BigInt(9007199254740993))).toBe("9007199254740993");
```
igdragoev: **This test always fails, and it's why Build & Test is red.** `BigInt(9007199254740993)` gets a number literal above `MAX_SAFE_INTEGER`, which rounds to `9007199254740992` before the conversion. Result: `expected 9007199254740992 to be 9007199254740993`. Use the BigInt literal `9007199254740993n`.

features/telemetry/infrastructure/telemetry-initializer.ts
```diff
@@ -1,98 +1,114 @@
...
 
-  /**
-   * Register handlers so uncaught exceptions and unhandled rejections are sent to AppInsights (Logs API with exception.* attributes).
-   */
-  private registerUnhandledErrorHandlers(): void {
-    process.on("uncaughtException", (err: Error) => {
-      TelemetryLogger.error("Uncaught exception", err, {}, "instrumentation");
-      this.sdk?.shutdown().finally(() => process.exit(1));
+  private registerCrashHandlers(sdk: TelemetrySdk): void {
+    process.once("uncaughtException", (error: Error) => {
```
igdragoev: **An uncaught exception now shuts down the Hub on Azure.** Crash handlers (flush, then `process.exit(1)`) are now registered in Azure mode. Before, the Azure strategy returned `null` and registered none, so Next.js's router-server logged the error and kept serving. With `APPLICATIONINSIGHTS_CONNECTION_STRING` set, any stray async throw (a socket `error` event, a library callback) now exits the process within ~5s and drops in-flight requests. OTLP already behaved this way; the PR extends it to Azure. Should telemetry setup decide whether the process lives or dies? A log-only handler would keep the old behaviour.

features/telemetry/infrastructure/telemetry-logger.ts
```diff
@@ -223,81 +98,70 @@ export class TelemetryLogger {
...
 }
 
-function sanitizeConsoleAttributes(attributes: LogAttributes): LogAttributes {
-  return Object.fromEntries(
-    Object.entries(attributes).map(([key, value]) => [
-      key,
-      sensitiveAttributePattern.test(key) ? "[REDACTED]" : value,
-    ]),
+function shouldPrintFormattedConsole(): boolean {
+  if (TelemetryConfig.hasActiveLogExporter()) {
```
igdragoev: **If the SDK fails to start, every log is silently dropped.** Whether to fall back to the console is decided from env config (`hasActiveLogExporter` / `hasActiveExporter`), not from whether the SDK actually started. Example: `APPLICATIONINSIGHTS_CONNECTION_STRING` has a typo → the `AzureMonitorTraceExporter` constructor throws `Invalid instrumentation key` → the initializer catches it, `logs.getLogger()` is a no-op, and the console fallback is still suppressed, even with `TELEMETRY_CONSOLE_FALLBACK=true`. Every `TelemetryLogger` call disappears in production. Suggest deciding this from a "started successfully" flag the initializer sets.

features/telemetry/infrastructure/telemetry-sdk.ts
```diff
@@ -0,0 +1,206 @@
...
+      !exporters.azureConnectionString &&
+      !exporters.otlpTracesProtocol &&
+      !exporters.otlpLogsProtocol
+    ) {
+      throw new Error(
+        "No telemetry exporter configured. Set APPLICATIONINSIGHTS_CONNECTION_STRING and/or OTEL_EXPORTER_OTLP_ENDPOINT.",
+      );
+    }
+
+    const spanProcessors = spanProcessorsFor(exporters);
```
igdragoev: **One bad exporter stops every destination from starting.** Exporters for all destinations are built in one unguarded pass. If `APPLICATIONINSIGHTS_CONNECTION_STRING` is malformed and a working `OTEL_EXPORTER_OTLP_ENDPOINT` is also set, the Azure exporter throws in `spanProcessorsFor` → no `NodeSDK`, and no OTLP traces or logs either. That contradicts the "fans out" contract. Suggest building each destination in its own try/catch, logging the failure and skipping only that one.

features/telemetry/infrastructure/telemetry-sdk.ts
```diff
@@ -0,0 +1,206 @@
...
+
+const OTLP_SPAN_PROCESSOR_OPTIONS = {
+  scheduledDelayMillis: 1000,
+  maxQueueSize: 2048,
+  maxExportBatchSize: 512,
+} as const;
+
+/** Exporters read OTEL_EXPORTER_OTLP_* themselves (endpoint, headers, TLS). */
+const OTLP_TRACE_EXPORTERS: Record<OtlpProtocol, () => SpanExporter> = {
+  grpc: () => new OTLPGrpcTraceExporter(),
```
igdragoev: **Existing plaintext gRPC collectors will break.** The gRPC exporters are now built without credentials. The old code forced `createInsecure()` for any endpoint not starting with `https://`. `otlp-grpc-exporter-base` requires TLS for endpoints without a scheme unless `OTEL_EXPORTER_OTLP_INSECURE=true`. A deployment with `OTEL_EXPORTER_OTLP_ENDPOINT=otel-collector:4317` (plaintext collector) will try TLS after upgrading, fail the handshake, and silently stop exporting traces and logs. Suggest keeping the old default, or documenting it as a breaking change.

features/telemetry/infrastructure/telemetry-sdk.ts
```diff
@@ -0,0 +1,206 @@
...
+      resource,
+      autoDetectResources: false,
+      spanProcessors,
+      metricReaders: [],
+      logRecordProcessors: [],
+      contextManager: new AsyncLocalStorageContextManager(),
+      sampler: new AlwaysOnSampler(),
+      instrumentations: [
+        new HttpInstrumentation(),
+        new UndiciInstrumentation(),
```
igdragoev: **Each Hub→API fetch shows up as two dependencies.** `UndiciInstrumentation` is now active on the Azure path, while Next.js still emits its own `AppRender.fetch` CLIENT span (it's in `NextVanillaSpanAllowlist`). A server action that calls the API produces two `dependencies` rows: Next's `fetch GET https://api…` plus the undici span. That doubles dependency volume and clutters the Application Map and transaction view.

features/telemetry/infrastructure/telemetry-initializer.ts
```diff
@@ -1,98 +1,114 @@
...
+      );
+      sdk
+        .forceFlush()
+        .catch((error) => console.error("Error flushing telemetry", error));
     };
 
-    // Register for various termination signals
-    process.on("SIGTERM", shutdownHandler);
-    process.on("SIGINT", shutdownHandler);
+    process.once("SIGTERM", () => flush("SIGTERM"));
```
igdragoev: **Spans and logs from the last requests are lost on shutdown.** The SIGTERM/SIGINT flush runs at the start of Next's cleanup, *before* `server.close()` drains in-flight requests. Nothing flushes after the drain, and Next then calls `process.exit(143)`. Spans and logs from requests that finish during the drain stay queued in `BatchSpanProcessor` (5s default delay) / `BatchLogRecordProcessor` and are lost, including the error logs from those final requests. The existing handler had this problem already; the PR extends it to Azure.

features/telemetry/infrastructure/redact-sensitive-attributes.ts
```diff
@@ -0,0 +1,16 @@
+const SENSITIVE_ATTRIBUTE_KEY =
+  /(authorization|cookie|token|secret|password|api[-_ ]?key|connection[-_ ]?string)/i;
```
igdragoev: **Redaction now removes non-secret diagnostic values.** Key-regex redaction used to apply only to the console fallback; it now runs before every exporter. It replaces any attribute whose key merely *contains* `token` etc. Example: `activate-invite.action.ts:63` logs `{ hasToken: true }` → App Insights and OTLP get `hasToken: "[REDACTED]"`. The boolean becomes a string, and the diagnostic the log exists for is lost. Suggest exact or suffix key matching, or skipping boolean/number values.

features/pdf-export/submission/render-submission-pdf.use-case.tsx
```diff
@@ -136,13 +146,45 @@ export async function renderSubmissionPdf({
...
     }
 
-    return Result.error("PDF export failed.");
+    // Safe scalars only: renderer errors can quote image URLs, and those may
```
igdragoev: **The SAS token still reaches telemetry through the span.** This "safe scalars only" log keeps SAS-token URLs out of `TelemetryLogger`. But `render-pdf` rethrows, so `TelemetryTracer.traceAsync` calls `span.recordException(error)` and exports the full error message to the same backends. A renderer error that quotes an image URL with `?sig=…` still lands in App Insights `exceptions` / OTLP. The span-side leak isn't new, but this PR now sends span exceptions to Azure the same way. The redaction here only covers half of it.

features/telemetry/infrastructure/telemetry-initializer.ts
```diff
@@ -1,98 +1,114 @@
...
 
     process.on("unhandledRejection", (reason: unknown) => {
-      const error =
-        reason instanceof Error
-          ? reason
-          : new Error(String(reason));
       TelemetryLogger.error(
         "Unhandled promise rejection",
-        error,
+        reason instanceof Error ? reason : new Error(String(reason)),
```
igdragoev: **Non-Error rejections log as `[object Object]`.** This wraps non-Error rejection reasons with `new Error(String(reason))`, which bypasses the `parseErrorMessage` / `stringifyUnknown` handling `TelemetryLogger.logFailure` already does. `Promise.reject({ code: 'E_API', message: 'boom' })` gets logged as `Unhandled promise rejection: [object Object]` instead of `boom`. Pass `reason` straight through.

features/telemetry/__tests__/telemetry-sdk.test.ts
```diff
@@ -0,0 +1,325 @@
...
+
+      const { strategy } = initialize();
+
+      expect(strategy.name).toBe("Azure AppInsights");
+      expect(traceExporters(strategy)).toHaveLength(1);
+      expect(logExporters(strategy)).toHaveLength(1);
+    });
+
+    it("fans out to Azure and OTLP when both are set", () => {
+      vi.stubEnv("APPLICATIONINSIGHTS_CONNECTION_STRING", AZURE);
```
igdragoev: **Setup and teardown don't match.** Each `initialize()` builds a real `NodeSDK`. `HttpInstrumentation` / `UndiciInstrumentation` enable themselves in their constructors (patching `http`, subscribing to `diagnostics_channel`), and real gRPC exporters and `BatchSpanProcessor`s get created. `afterEach` only calls `logs.disable()`. About 14 `initialize()` calls leave stacked http patches and undici subscriptions active in the vitest worker, so later fetch/http tests in the same worker create spans through leftover processors. Expect flaky or slow runs and open handles. Suggest shutting the SDK down in `afterEach`, or mocking `NodeSDK`/the instrumentations.

features/telemetry/infrastructure/telemetry-sdk.ts
```diff
@@ -0,0 +1,206 @@
...
+      ],
+    });
+
+    logs.setGlobalLoggerProvider(loggerProvider);
```
igdragoev: **Logging only works because of call order.** `NodeSDK` with `logRecordProcessors: []` still builds and registers an empty `LoggerProvider` in `start()`. Ours stays active only because the first registration wins. If a refactor calls `logs.setGlobalLoggerProvider` after `sdk.start()`, the empty provider wins and every `TelemetryLogger.emit` becomes a silent no-op. Passing the processors to `NodeSDK`, or skipping its provider, removes that risk and the redundant second provider.

features/telemetry/__tests__/telemetry-sdk.test.ts
```diff
@@ -0,0 +1,325 @@
...
+      expect(logExporters(strategy)).toHaveLength(2);
+    });
+
+    it("defaults OTLP to gRPC when no protocol is set", () => {
+      vi.stubEnv("OTEL_EXPORTER_OTLP_ENDPOINT", OTLP_GRPC);
+
+      const { strategy } = initialize();
+
+      expect(strategy.name).toBe("OTel");
+      expect(traceExporters(strategy)).toEqual([
```
igdragoev: **Missing AAA labels.** New tests with separate setup, call and expect phases (e.g. `wires Azure trace and log exporters`, most of `telemetry-config.test.ts` and `string-utils.test.ts`) have no `// Arrange` / `// Act` / `// Assert` labels, which `AGENTS.md` requires.

features/telemetry/infrastructure/filtering-span-processor.ts
PeterMarinov: Fixed in d5eea364. Noise is matched on `url.path`/`url.query` plus legacy `http.target`/`http.url`, and dropped for the whole trace (Next's request span has wrapper parents and render children with no URL). Sampling: Azure's own `RateLimitedSampler`, 5/s by default with Azure, configurable via `TELEMETRY_TRACES_PER_SECOND` or standard `OTEL_TRACES_SAMPLER`.

lib/utils/__tests__/string-utils.test.ts
PeterMarinov: Fixed in d5eea364: `BigInt("9007199254740993")` (the `n` literal is rejected by our TS target).

features/telemetry/infrastructure/telemetry-initializer.ts
PeterMarinov: Agreed. Fixed in d5eea364: uncaught exceptions and rejections are only logged. No shutdown, no exit; Next keeps serving.

features/telemetry/infrastructure/telemetry-logger.ts
PeterMarinov: Fixed in d5eea364: `TelemetryRuntime` records whether a log pipeline actually started. If none runs, `TelemetryLogger` prints to the console when log export was configured, fallback is on, or in dev.

features/telemetry/infrastructure/telemetry-sdk.ts
PeterMarinov: Fixed in d5eea364: each exporter is built in its own try/catch, logged and skipped on failure. Startup throws only if none can be built.

features/telemetry/infrastructure/telemetry-sdk.ts
PeterMarinov: Fixed in d5eea364: `http://` and scheme-less `host:port` stay plaintext by default. `OTEL_EXPORTER_OTLP_[SIGNAL_]INSECURE` overrides; `https://` is always TLS.

features/telemetry/infrastructure/telemetry-sdk.ts
PeterMarinov: Undici has to stay: Next's fetch span doesn't inject `traceparent`, so without it Hub -> API traces break. Fixed the duplicate in d5eea364 by defaulting `NEXT_OTEL_FETCH_DISABLED=1` (respected if set explicitly). One dependency per fetch, verified end to end.

features/telemetry/infrastructure/telemetry-initializer.ts
PeterMarinov: Fixed in d5eea364: after SIGTERM/SIGINT, `ExitFlush` holds Next's post-drain `process.exit(143|130)` until the SDK shuts down (max 5s). `process.exit` is untouched before a signal.

features/telemetry/infrastructure/redact-sensitive-attributes.ts
PeterMarinov: Fixed in d5eea364: booleans and numbers under credential-like keys are kept, so `hasToken: true` stays.

features/pdf-export/submission/render-submission-pdf.use-case.tsx
PeterMarinov: Fixed in d5eea364: `TelemetryTracer` redacts recorded exceptions, and the span filter redacts secret query params (`sig`, S3/GCS signatures, `token`, `code`) in URL attributes, span names and exception events for every span.

features/telemetry/infrastructure/telemetry-initializer.ts
PeterMarinov: Fixed in d5eea364: the reason is passed through, so `logFailure` describes it (`boom`, not `[object Object]`).

features/telemetry/__tests__/telemetry-sdk.test.ts
PeterMarinov: Fixed in d5eea364: instrumentations are faked in the SDK tests, and every created SDK is shut down in `afterEach`.

features/telemetry/infrastructure/telemetry-sdk.ts
PeterMarinov: Fixed in d5eea364: processors are passed to `NodeSDK`, which owns the only LoggerProvider. No second provider, no ordering dependency.

features/telemetry/__tests__/telemetry-sdk.test.ts
PeterMarinov: Fixed in d5eea364: AAA labels added across the rewritten SDK, initializer, logger, config and filter tests.

features/telemetry/infrastructure/filtering-span-processor.ts
PeterMarinov: Follow-up in 8e2c27f9: RSC requests (`_rsc`) are now traced, not dropped. They're real navigations that render and call the API, and sampling bounds their volume. Static files served by the router, which only produce URL-less `NextServer.*RequestHandler` wrapper spans, now export nothing.

igdragoev (approved):
All review comments addressed — thanks.

